PR #6, opened 2020-10-05
Updating docs to reflect implementation requirements.

stupergenius:
Update docs to replace `records` in certain places with `record`. The implementation expects `record` (singular), so updated the documentation to match.


phenax (approved):
Thank you

PR #12, opened 2020-10-06
added abhinavdalal.json

abhinavdalal:


domains/abhinavdalal.json
```diff
@@ -0,0 +1,12 @@
+{
+  "description": "Abhinav Dalal's website",
+  "repo": "https://github.com/b-om/b-om.github.io",
+  "owner": {
+    "username": "abhinavdalal",
+    "email": "abhinavdalal@gmail.com"
+  },
+  "record": {
+    "CNAME": "https://b-om.github.io/who-we-are.html",
```
phenax, line 9: CNAME has to be a domain name. URLs won't work. You should create a `abhinavdalal.github.io` repo and place the html there. Or you can redirect to the url using just `"URL": "https://b-om.github.io/who-we-are.html"`

phenax (approved):

domains/abhinavdalal.json
phenax, line 9: CNAME has to be a domain name. URLs won't work. You should create a `abhinavdalal.github.io` repo and place the html there. Or you can redirect to the url using just `"URL": "https://b-om.github.io/who-we-are.html"`

phenax (changes requested):

domains/abhinavdalal.json
abhinavdalal: fixed cname

phenax (approved):

PR #40, opened 2020-10-07
Create anthony.json

hubsta:


domains/anthony.json
```diff
@@ -0,0 +1,12 @@
+{
+  "description": "Anthony's personal developer website",
+  "repo": "https://github.com/github-username/github-username.github.io",
```
phenax, line 3: If you are using github pages to host the website, please correct the link. Otherwise, you can remove the `repo` from the json file.

phenax (changes requested):

domains/anthony.json
hubsta: Removed that, thanks.

phenax (approved):

PR #42, opened 2020-10-07
Registered for my domain

danhab99:


domains/dan-habot.is-a.dev
```diff
@@ -0,0 +1,12 @@
+{
```
phenax, line 1: The file name is incorrect. You need to rename it to `dan-habot.json`

phenax (changes requested):

phenax (approved):

PR #54, opened 2020-10-08
Create andros.json

tanrax:


domains/andros.json
```diff
@@ -0,0 +1,11 @@
+{
+  "description": "Andros's personal developer website",
+  "repo": "https://github.com/tanrax",
+  "owner": {
+    "username": "Andros Fenollosa Hurtado",
+    "email": ""
```
phenax, line 6: If you don't wish to share your email address, you should add another key below with a link to any of your socials so that we can get in touch with you just in case.

phenax (changes requested):

domains/andros.json
tanrax: All right! I make the change. Thanks for the initiative :)

tanrax:
Would this be correct?

phenax (approved):

tanrax:
great! Is it possible to add the #hacktoberfest-accepted tag?

PR #66, opened 2020-10-08
add alexlykesas.is-a.dev domain

alexookah:


domains/alexlykesas.json
```diff
@@ -0,0 +1,12 @@
+{
+    "description": "Alex Lykesas's personal developer website",
+    "repo": "https://github.com/Alexookah/alexlykesas.com",
+    "owner": {
+        "username": "alexookah",
+        "email": "alexlykesas@gmail.com"
+    },
+    "record": {
+        "CNAME": "alexlykesas.com",
+        "URL": "https://alexlykesas.is-a.dev"
```
phenax, line 10: If you have a dot com, I'd suggest just using this subdomain to redirect to your website instead of using a cname record. You'll have to configure a lot less stuff.

Something like,
```
"URL": "https://alexlykesas.com"
```

If you want to keep using the cname, let me know. I'll merge it.

Alexookah: Just updated the record as you suggested :)

PR #83, opened 2020-10-09
create ezedinff.json

ezedinff:


phenax (approved):

phenax:
@ezedinff, The above comment doesn't mention the steps required for vercel but it should be pretty similar to setup a custom domain for your app. I found [this](https://vercel.com/docs/custom-domains) but I'm not sure if it is the right one. You can create a support issue with a link to this pr if you have any trouble setting things up.

PR #153, opened 2020-10-09
Create ashleymavericks.json

ashleymavericks:


domains/ashleymavericks.json
```diff
@@ -0,0 +1,11 @@
+{
+  "description": "Portfolio Website",
+  "repo": "https://github.com/ashleymavericks/Portfolio-Site",
+  "owner": {
+    "username": "ashlyemavericks",
+    "email": "singhanurag024@gmail.com"
+  },
+  "record": {
+    "URL": "https://anuragsingh.dev"
```
phenax, line 9: We are currently working on a bug related to url redirections #221. This may take a week to fix. I'd suggest using a cname record instead. If you want to keep it as it is, let me know, I'll merge it in.

phenax (changes requested):

domains/ashleymavericks.json
phenax: This issue has been fixed. Will merge the pr now

phenax (approved):

PR #155, opened 2020-10-09
Create atlas.json

robertdrakedennis:


phenax (approved):

phenax:
The above reply doesn't mention the steps for netlify but I found [this doc](https://docs.netlify.com/domains-https/custom-domains/) that might help.

PR #157, opened 2020-10-09
Create nishant.json

nishch:


domains/nishant.json
```diff
@@ -0,0 +1,11 @@
+{
+  "description": "Nishant's website and blog",
+  "repo": "https://github.com/nishch/nishch.github.io",
+  "owner": {
+    "username": "nishch",
+    "email": "nishantchvedi@gmail.com"
+  },
+  "record": {
+    "URL": "https://twistedsoft.com"
```
phenax, line 9: We are currently working on a bug related to url redirections #221. This may take a week to fix. I'd suggest using a cname record instead. If you want to keep it as it is, let me know, I'll merge it in.

phenax (changes requested):

domains/nishant.json
nishch: @phenax thanks for the update, I have updated the PR with `CNAME` record.

phenax (approved):

phenax:
@nishch, The issue with redirection has been fixed. If you want, you can change this back to a URL redirection instead of CNAME.

PR #163, opened 2020-10-10
add sycured

sycured:


domains/sycured.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "sycured",
+        "email": "sycured@protonmail.com"
+    },
+    "record": {
+        "URL": "https://sycured.com"
```
phenax, line 7: We are currently working on a bug related to url redirections #221. This may take a week to fix. I'd suggest using a cname record instead. If you want to keep it as it is, let me know, I'll merge it in.

phenax (changes requested):

domains/sycured.json
sycured: I want to keep it as it is. Thanks

phenax: Alright. You can subscribe to #221 to get updates on the issue.

phenax (approved):

phenax:
@sycured The issue with redirection has been fixed. Feel free to create an issue if you have any trouble with your subdomain.

PR #183, opened 2020-10-10
Added sid.json

SiddPatki:


domains/sid.json
```diff
@@ -0,0 +1,12 @@
+{
+  "description": "Siddhartha Patki's development portfolio",
+  "repo": "https://github.com/SiddPatki",
+  "owner": {
+    "username": "SiddPatki",
+    "email": "siddharthapatki@gmail.com"
+  },
+  "record": {
+    "CNAME": "https://siddpatki.github.io/my-site/",
```
phenax, line 9: CNAME records cannot be a full url. It has to be a host name. I'd suggest you rename your repo to `siddpatki.github.io`.

You can take a look at the [github pages guide](https://guides.github.com/features/pages/).

phenax (changes requested):

domains/sid.json
SiddPatki: Sorry and thanks...I am a noob in this so was not sure.
I will update the CNAME as per your comment.
Will I be able to change this CNAME in the future if I want it to redirect to some different link

phenax: You can change the record any time you want with a new pull request

phenax (approved):

PR #193, opened 2020-10-11
Create sarath.json

srttk:
sarath.json added


phenax (approved):

domains/sarath.json
```diff
@@ -0,0 +1,12 @@
+{
```
phenax, line 1: That domain seems to have already been take unfortunately

phenax (changes requested):

PR #194, opened 2020-10-11
fix: adds CNAME to template json

shajanjp:


domains/@.json
```diff
@@ -6,6 +6,7 @@
     "email": "phenax5@gmail.com"
   },
   "record": {
-    "URL": "http://www.is-a.dev"
+    "URL": "http://www.is-a.dev",
+    "CNAME": "github-username.github.io"
```
phenax, line 10: This isn't the right file you're trying to edit there. This is for the root domain

phenax (changes requested):

PR #198, opened 2020-10-11
Create anwar.json

getanwar:


domains/anwar.json
```diff
@@ -0,0 +1,11 @@
+{
+    "description": "Anwar's personal website",
+    "repo": "https://github.com/getanwar",
+    "owner": {
+        "username": "getanwar",
+        "email": "say@anwar.im"
+    },
+    "record": {
+        "URL": "https://anwar.im"
```
phenax, line 9: We are currently working on a bug related to url redirections #221. This may take a week to fix. I'd suggest using a cname record instead for now. If you want to keep it as it is, let me know, I'll merge it in.

phenax (changes requested):

domains/anwar.json
getanwar: I don't mind if you merge it in.

phenax: The issue has been fixed. Will merge it.

phenax (approved):

PR #265, opened 2020-10-14
Create harikrishnanps

super-jedi:
Added a domain json file


domains/harikrishnanps
```diff
@@ -0,0 +1,12 @@
+
```
phenax, line 1: The file must be a `.json` file

phenax (changes requested):

PR #269, opened 2020-10-15
Create hari.json

harishnkr:


shajanjp:
Hi @Haritronic909 
It will be good if you can change the CNAME to lower case.

phenax:
> It will be good if you can change the CNAME to lower case.
Wrong cased cnames should be fine. I'm merging it in for now.

Also, @shajanjp, thanks a lot for helping out with these reviews! I really appreciate it.

phenax (approved):

Haritronic909:
thanks for helping me out, is the cname case sensitive?? Idk that's y I'm
asking

shajanjp:
@Harikrishnan6336 Its not case sensitive, but same cased would be better right ?
Anyway, your chages are merged.
Thanks for contributing. :+1:

PR #315, opened 2020-10-19
Create pratyay.json

X3NOSIZ:


phenax (approved):
Currently we haven't documented the exact steps required for configuring netlify. I have found this [doc](https://docs.netlify.com/domains-https/custom-domains/) that might help. Please let me know if the link helped so I can document it for other netlify users.

X3NOSIZ:
> 
> 
> Currently we haven't documented the exact steps required for configuring netlify. I have found this [doc](https://docs.netlify.com/domains-https/custom-domains/) that might help. Please let me know if the link helped so I can document it for other netlify users.

Thanks for the help but unfortunately it didn't work :(
Well I can do one thing, I can update the portfolio endpoint in the pr to `neilblaze.js.org`

phenax:
Best to use github pages as it is the simplest to configure. Not sure how `.js.org` will be configured as you will also need the ssl certificate to  include your new subdomain.

PR #330, opened 2020-10-20
Create riya.json

riya3000:
creating my first PR


phenax (changes requested):
This is an incorrect record. You should read through the [docs](https://github.com/is-a-dev/register/blob/main/API.md) for information about how to create a domain file. You can also look at other files in the domains directory for example. Let me know if you need more help with this.

PR #351, opened 2020-10-22
add rick

ricjd:
Thanks!!


phenax (changes requested):
The json file should be in the `domains` directory

PR #387, opened 2020-10-31
Create starkblaze01.json

starkblaze01:


domains/starkblaze01.json
```diff
@@ -0,0 +1,11 @@
+{
+  "description": "starkblaze01's world",
+  "repo": "https://github.com/starkblaze01/starkblaze01",
+  "owner": {
+    "username": "starkblaze01",
+    "email": "mp.pathela@gmail.com"
+  },
+  "record": {
+    "CNAME": "https://starkblaze01.netlify.app/"
```
phenax, line 9: CNAME has to be a hostname

```suggestion
    "CNAME": "starkblaze01.netlify.app"
```

phenax: Also, please confirm that you can configure custom domains with netlify

phenax (changes requested):

domains/starkblaze01.json
starkblaze01: Hi, will changing "CNAME" to URL for redirect will do?
Yes, we can configure custom domains with netlify.

phenax: If you want to change it to a URL redirection then you don't need to configure anything in netlify.

starkblaze01: Okay, thanks. Updated the patch.

phenax (approved):

PR #397, opened 2020-11-08
Update ysfchn.json

ysfchn:
I thought setting the record as CNAME would be efficient first, but looks like it is causing a problem on my end, so I would like to set it as URL forwarding, sorry for the inconvenience 😅


phenax (approved):
No worries

PR #421, opened 2020-11-23
Update george.json due to DNS error

aykarageorge:


phenax (changes requested):
This wont help. To fix the dns error you need to configure cloudflare. Alternatively, you can go for URL redirection instead of CNAME subdomain.

```
    "URL": "https://george.aykara4.com"
```

phenax (dismissed):

phenax (approved):

PR #448, opened 2020-12-11
Hello, I want to register a domain. Can I?

ghost:


domains/chucha.json
```diff
@@ -0,0 +1,10 @@
+{
+  "owner": {
+    "username": "Sly-Little-Fox",
+    "email": "paranormal.studio123@gmail.com"
+  },
+  }
```
phenax, line 6: Seems to be invalid json

phenax (changes requested):

PR #478, opened 2021-01-04
Delete whitekj.json

juna1221:


phenax:
@CwhiteKJ The last deployment had errored out which is why your domain was never published. Sorry about that. The domain should be available now unless cached. If you still want to delete the domain file tag me on this pr and I'll merge it in.

PR #504, opened 2021-01-16
Update piyush.json | Change redirect to default github pages

whopiyush:
My domain to which the service redirected is expired, I would like if it shall redirect to my github pages URL now :p

### DIFF
--- a/domains/piyush.json
+++ b/domains/piyush.json
@@ -5,6 +5,6 @@
     "email": "bhangalepiyush@gmail.com"
   },
   "record": {
-    "URL": "https://piyush.codes"
+    "URL": "https://officialpiyush.github.io"
   }
 }


domains/piyush.json
```diff
@@ -5,6 +5,6 @@
     "email": "bhangalepiyush@gmail.com"
   },
   "record": {
-    "URL": "https://piyush.codes"
+    "URL": "https://officialpiyush.github.io"
```
phenax, line 8: @officialpiyush If you want it to be a sub-domain instead of a URL redirection, you can use a CNAME record instead and add the sub-domain to your repo settings.

```suggestion
    "CNAME": "officialpiyush.github.io"
```

Let me know if you want it to keep it as is, I'll merge it in.

officialpiyush: I would like to keep this as it is :)

phenax (approved):

PR #540, opened 2021-02-01
Create b.json

bashbaugh:


phenax (approved):

phenax (changes requested):
Oops. Missed that. The subdomain has to be more than 2 characters long as mentioned in the [doc](https://github.com/is-a-dev/register/blob/main/docs/domains-json.md#filename)

PR #554, opened 2021-02-06
Create shreyy.json

tejasag:
Making this for a friend as he is very new to git and github overall and just made his first website today :)
I hope u understand


phenax (approved):

phenax (changes requested):
You need to specify either an email or twitter or some other social media contact. Also, the repo doesn't seem to exist at the moment. I'd suggest you at least create a github pages repo with a single index.html file so that configuration for the site after the record gets merged becomes easier.

PR #594, opened 2021-02-27
JSON, branch and API description update

sshyran:
As  is-a.dev itself is served through Github Pages, several descriptions changed in regard to this.

### DIFF
--- a/API.md
+++ b/API.md
@@ -11,20 +11,21 @@
 ```json
 {
   "description": "Add some description",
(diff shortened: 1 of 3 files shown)


phenax (changes requested):
Most of the changes made here are incorrect. The reason your domain is returning a 404 is because you haven't added your domain name to the repo settings as I can't see a `CNAME` file in your repo. You can keep the changes to your domain file and undo the other changes.

PR #603, opened 2021-03-04
Adding a my jetus.is.a.dev

j3tus:


domains/jetus.is.a.dev
```diff
@@ -0,0 +1,12 @@
+{
+  "description": "A personal is a dev site.",
+  "repo": "https://github.com/Dev-Jetus/register",
```
phenax, line 3: This should be the repo hosting your github pages site

domains/jetus.is.a.dev
```diff
@@ -0,0 +1,12 @@
...
+  "owner": {
+    "username": "jetus",
+    "email": "officialjetus@gmail.com",
+    "twitter": "itzjetus"
+  },
+  "record": {
+    "CNAME": "https://github.com/dev-jetus/dev-jetus.github.io"
```
phenax, line 10: `CNAME` has to be a hostname. In this case, I think you meant `dev-jetus.github.io`

phenax (changes requested):

domains/jetus.is.a.dev
Dev-Jetus: Can i just leave this for a bit, and i will do this later on.

TheBotlyNoob: 🤦‍♂️ Just change the CNAME to dev-jetus.github.io then you can get the subdomain...

PR #651, opened 2021-04-02
Update raksix.json

Raksixoffical:


TheBotlyNoob (changes requested):
close this PR please, the IP doesn't work, and you have #652 going as well (and that doesn't work either)

PR #689, opened 2021-04-13
Form completing for is-a-dev

didiermun:


hackermondev (changes requested):
The CNAME record is not supposed to be the custom domain of your webpage. (If you use github pages, then use the github.io link)

didiermunezero:
> The CNAME record is not supposed to be the custom domain of your webpage. (If you use github pages, then use the github.io link)

ok gonna update and i will open again

PR #730, opened 2021-04-23
Setup NodeJS CI

Craftzman7:
# NodeJS CI
This will add NodeJS CI. All this does is run ```npm build``` and ```npm ci``` to check for a malformed package.json and any invalid packages


hackermondev:
e

Craftzman7:
> 
> 
> e

e

phenax (commented):
Instead of creating a new workflow, we should update the existing `checks` and `publish-records` workflow. Also, we can avoid using a matrix of versions as our runtime target is the github action itself so it won't add any value.

PR #815, opened 2021-05-28
updating killua.json

Killua7362:


TheBotlyNoob:
You cannot have capital letters in the filename

PR #958, opened 2021-07-23
Update mert.json

mertcelik2196:


hackermondev (changes requested):
You cannot have CNAME and URL

PR #969, opened 2021-07-26
Create lthon09.json

lthon09:


hackermondev (changes requested):
CNAME cannot be a link. Did you mean to put ``URL`` redirection instead?

PR #1022, opened 2021-08-08
updated joker.json (CNAME)

JoKeRxTD:


TheBotlyNoob (changes requested):
The `CNAME` cannot be an array

mtgsquad:
Hello, I don't think you understand how is-a.dev provides subdomains, it is a part of our terms that we do not allow nameservers on subdomains, however you can set your CNAME to a string and then you can use it with cloudflare pages & etc.

JoKeRxTD:
> Hello, I don't think you understand how is-a.dev provides subdomains, it is a part of our terms that we do not allow nameservers on subdomains, however you can set your CNAME to a string and then you can use it with cloudflare pages & etc.

Hmm I have the Webspace. Just need the domain. I will have to google it. The IP I placed before is the Shared IP for my webhosting ns1.gtxgaming.co.uk is the nameserver for it

Danix421:
> > Hello, I don't think you understand how is-a.dev provides subdomains, it is a part of our terms that we do not allow nameservers on subdomains, however you can set your CNAME to a string and then you can use it with cloudflare pages & etc.
> 
> Hmm I have the Webspace. Just need the domain. I will have to google it. The IP I placed before is the Shared IP for my webhosting ns1.gtxgaming.co.uk is the nameserver for it

I'll repeat what @mtgsquad said: is-a.dev doesn't allow namespaces, so you can't use them (I think).

JoKeRxTD:
Updated.

hackermondev (approved):

TheBotlyNoob (approved):

PR #1060, opened 2021-08-22
Remove incorrect subdomains

hackermondev:
Remove flash subdomain. It was added to the wrong directory. It is supposed to be in the ``domains`` folder but it is in the root directory.


hackermondev:
also remove sociopath domain

TheBotlyNoob (approved):

mtgsquad (approved):
Seems good.

PR #1065, opened 2021-08-24
add a missing word

AlmostSuspense:
because one word makes all the difference


hackermondev (approved):
thanks for catching this!

mtgsquad (approved):

PR #1139, opened 2021-09-12
move to zoho mail

TheBotlyNoob:


mtgsquad (approved):

RehanPlayz:
Will I be able to send emails from the subdomain if I do something like this

RehanPlayz (approved):

hackermondev (approved):

TheBotlyNoob:
@RehanPlayz yep! see https://github.com/is-a-dev/register/blob/main/docs/domains-json.md#record-required

RehanPlayz:
> @RehanPlayz yep! see https://github.com/is-a-dev/register/blob/main/docs/domains-json.md#record-required

Ok so do I have to make two PRs because one for the Domain Validation and the other for the actual  MX records and what's the difference between Zoho and ImprovMX (I have only used ImprovMX before) is there any advantage of one over the other

PR #1144, opened 2021-09-13
Rename rehan-playz.json to rehanplayz.json

RehanPlayz:
and added a TXT record for Zoho Mail domain verification


phenax (changes requested):
You can't have a CNAME and a TXT record in conjunction. Maybe try using `A` record if possible.

PR #1151, opened 2021-09-15
Switched to URL redirection

Radiergummi:
@TheBotlyNoob initially wanted to use a `CNAME` but that didn't work for various reasons, then got work stuff in between. Thank you for reminding us of broken domains :) Happy to have a is-a.dev


TheBotlyNoob (approved):
> Thank you for reminding us of broken domains

no problem!

hackermondev (approved):

PR #1175, opened 2021-09-19
Added subdomains files

MinMitCoder:
I'm using 2 subdomains because CNAME records cant go with MX and TXT records.
Hope it's allowed!


hackermondev (approved):
yep, its allowed!

TheBotlyNoob (approved):

PR #1335, opened 2021-11-01
Create anon-cdn.json

anondev-sudo:
I would like to get connect this to tixte uploader.

They require you to add TXT record, is it possible?

image


phenax:
Sorry for the delayed response. Unfortunately, we cannot allow generic challenges like that since it allows verifying domains that may not be yours.

PR #1352, opened 2021-11-10
Create JavaJumper.json

JumperOnJava:


phenax (changes requested):
There are a few things missing from your record file. Please refer to the docs or you can also check out existing records for reference

PR #1354, opened 2021-11-11
yieldray.json

YieldRay:


TheBotlyNoob (changes requested):
Netlify doesn't work with is-a.dev, use [Railway](https://railway.app?referralCode=jj) instead

PR #1375, opened 2021-11-18
Change mehf.json (cname error)

dzired:
I think I made an error with CNAME and site is not letting me enforce HTTPS, this might be a fix

"Unavailable for your site because your domain is not properly configured to support HTTPS"

### DIFF
--- a/domains/mehf.json
+++ b/domains/mehf.json
@@ -6,6 +6,6 @@
     "email": "mehf@protonmail.com"
   },
   "record": {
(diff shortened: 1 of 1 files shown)


dzired:
Was having troubles since I wanted .is-a.dev to be the main domain and I think I had it so the domain would redirect to my CNAME instead of being my main domain. This should be right now I think

hackermondev (approved):

PR #1379, opened 2021-11-18
Create comodino.json

Comodinoh:


phenax (changes requested):
You have changes here unrelated to your record. You'll have to create a new PR with just the `domains/comodino.json`file

PR #1392, opened 2021-11-20
Create nebulabc.json

NebulaBC:


phenax (changes requested):
You are missing a few fields. You could check the documentation or existing records for reference

PR #1409, opened 2021-11-27
Update aditya.json

a69y:
Added MX Records, Removed CNAME Because It Doesn't Works With MX, Added A Records For GH Pages!.

Reviewer, Please See That All Things Are Right!! :)


icancodebeauty:
@TheBotlyNoob & @mtgsquad Please Review This :)

TheBotlyNoob (approved):

PR #1420, opened 2021-11-29
Create devin.json

iDrizzle:
Add this!


phenax (changes requested):
There is a syntax error in your record file. Verify your json or you could also look at existing domains for reference

PR #1433, opened 2021-12-03
Rename roger.json to zephyr.json, update info

UsrBinLuna:


phenax (changes requested):
You can't remove a record that was created from a different account. Please use the `rogerpanza` account to approve this PR.

PR #1436, opened 2021-12-04
Create dinothedev.json

ghost:
For me dinothedev

### DIFF
--- a/domains/dinothedev.json
+++ b/domains/dinothedev.json
@@ -6,6 +6,6 @@
     "email": "dinothedevofficial@gmail.com"
   },
   "record": {
-    "CNAME": "https://dinothedevofficial.repl.co"
+    "CNAME": "dinothedevofficial.repl.co"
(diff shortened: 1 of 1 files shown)


phenax (changes requested):
If you meant to use URL redirection, use URL instead of CNAME. If you meant to use a subdomain, please read repl.co's documentation regarding custom subdomains.

DinoTheDevOfficial:
I know how replits domain thing works you need to use a cname pointing to a replit. if you create a replit the same as your username it assigns the replit the <username>.repl.co domain

phenax:
@DinoTheDevOfficial, CNAME's are hostnames not urls. As you may have seen in the repl it docs, there is no `https://` in the CNAME so in this case you'll need to remove it

DinoTheDevOfficial:
fixed

phenax (approved):

PR #1437, opened 2021-12-04
Create jaikt.json

jaikt:


phenax (changes requested):
You have a syntax error in your domain json file. You can use an existing record for reference.

PR #1453, opened 2021-12-08
dexo.json

DeXoHigh:
hi, this is for my website


phenax (changes requested):
Unfortunately, you cannot use CNAME in conjunction with other records. This is a limitation in DNS itself. You can use A records instead of CNAME.

PR #1455, opened 2021-12-09
Create py.json

drugtraffic:
hi, this is for my website


phenax (changes requested):
CNAME has to be a hostname. It cannot be a full URL

PR #1464, opened 2021-12-12
create benjo.json

Golgrax:


phenax (changes requested):
You have a syntax error in your domain json file

PR #1467, opened 2021-12-12
Proximity

hexed:


domains/proximity.json
```diff
@@ -0,0 +1,9 @@
+{
+  "description": "Dev Env",
+  "repo": "https://github.com/hexed",
+  "owner": {
+    "username": "hexed",
+    "email": "me@proximity.sh"
+  },
+  "record": {"A": "144.172.67.6"}
```
phenax, line 8: A records have to be an array of ips

phenax (changes requested):

domains/proximity.json
hexed: Why an array ?  Only need the one up

phenax: The schema as described in the docs requires that the value be an array since technically, "A" could have multiple values. Please change wrap the ip in an array and I'll merge it in.

hexed: @phenax Got it.

phenax (approved):

PR #1469, opened 2021-12-13
Create nkno.json

ghost:


phenax (changes requested):
Please use your github username instead of `github-username` and your own email address in your record file

PR #1476, opened 2021-12-14
Create rlx.json

RLX-OP:


phenax (changes requested):
Need an email address to contact you in case you need to be notified of any changes to is-a-dev. Alternatively, you can keep email as "" and add any of twitter, discord, etc.

RLX-OP:
@phenax see latest one bcz closing this

PR #1479, opened 2021-12-16
Create devilnator.json

hexaorzo:


phenax (changes requested):
We do not support AAAA records yet. Use A is possible. If not, you can create a feature request issue for AAAA record support.

PR #1496, opened 2021-12-20
add joshua.json

Joshua-Noakes1:
adding joshua.is-a.dev


phenax (changes requested):
`A` record has to be an array of ips

Joshua-Noakes1:
Added an array!

phenax (approved):

PR #1505, opened 2021-12-23
create test.r2d2.json

hogakesan:


phenax (changes requested):
Not sure if this is what you intended but you need `r2d2.is-a.dev` to be the target cname. Please use the cname of the service where the page is hosted.

PR #1531, opened 2022-01-05
Update scientific-dev.json

scientific-dev:
I bought a new domain https://thesudarsandev.xyz/ so I thought to set a redirect instead of a CNAME.
Are the following changes correct for a redirect?


scientific-dev:
When will this be merged?

phenax (approved):

PR #1552, opened 2022-01-13
Create azer.json

Sadykhzadeh:


phenax (changes requested):

Sadykhzadeh:
> If you want a URL redirection, use URL instead of CNAME.

Thank you a lot! Changed it as you said. Please, review this PR again :3

phenax (approved):

PR #1563, opened 2022-01-16
Applying for "stringent.is-a.dev" subdomain

ghost:
I am currently building the portfolio and it's backend using vuetify and nuxt so please bare with me.


StringentDev:
Just added some mobile support to the full extent. Vue is awesome; I can just easily get an app to look how I want it to look

phenax (approved):

PR #1586, opened 2022-01-20
a-furry.json

Sxshley:
ignore the typh.json, i wasnt paying attention when i was making it


phenax (changes requested):
Unfortunately, netlify doesn't support is-a-dev cnames. You can use URL redirection instead or a different service like railway.app

phenax (approved):

PR #1589, opened 2022-01-21
Add jalaj.json

jalajcodes:


phenax (changes requested):
Unfortunately, netlify doesn't support is-a-dev cnames. You can use URL redirection or a different hosting service like railway.app

jalajcodes:
> Unfortunately, netlify doesn't support is-a-dev cnames. You can use URL redirection or a different hosting service like railway.app

@phenax Changed to Vercel

jalajcodes:
Nvm just found that vercel is also not supported.

Still figuring out how to deploy gatsby site to railway.

phenax (approved):

PR #1591, opened 2022-01-21
add nuxsh.json

nuxshed:


phenax (changes requested):

nuxshed:
oops. im sorry for that

fixed it now @phenax

phenax (approved):

PR #1634, opened 2022-01-28
changed CNAME

Felga7453:
Changed CNAME because it was a wrong site


phenax (changes requested):

phenax:
Closing in favor of #1635

PR #1654, opened 2022-02-03
Create cornix

norahlol:


phenax (changes requested):
Due to limitations in DNS, `CNAME` records cannot be used in conjunction with any other records.
Also, please add the `.json` extension to your file

MysteryDev01:
Ok thanks. i will do that

PR #1660, opened 2022-02-04
Add: domains/zuhaib.json

powrhouseofthecell:


phenax (changes requested):
You cannot use the 3 records together. Please read the docs and have a look at existing domain records for reference.

phenax (changes requested):
We need an email address to contact you in case you need to be notified of any changes to is-a-dev. Alternatively, you can keep email as "" and add any of twitter, discord, etc.

phenax (dismissed):

phenax (approved):

PR #1666, opened 2022-02-05
Create rstar.json

Rstar284:


phenax (changes requested):
There is a json syntax error in your domain file

PR #1690, opened 2022-02-10
Fix some outdated info.

MaskDuck:


phenax (changes requested):

aduckuwu:
Fixed =))

TheBotlyNoob (dismissed):
that is not a very popular opinion ngl

image

TheBotlyNoob (approved):

phenax (approved):

PR #1697, opened 2022-02-12
created herkura.json file

herkura:
I have not included my email. Thanks:)

### DIFF
--- a/domains/herkura.json
+++ b/domains/herkura.json
@@ -1,0 +1,10 @@
+{
+  "description": "Herkura's Portfolio",
+  "repo": "https://github.com/herkura/.me",
+  "owner": {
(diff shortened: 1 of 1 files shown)


domains/herkura.json
```diff
@@ -0,0 +1,10 @@
+{
+  "description": "Herkura's Portfolio",
+  "repo": "https://github.com/herkura/.me",
+  "owner": {
+    "username": "herkura",
```
phenax, line 5: We need a way to contact you in case you need to be notified of any changes to your domain or the is-a-dev service. Alternatively, you can keep email as "" and add an active account username for twitter, discord, etc.

phenax (changes requested):

PR #1703, opened 2022-02-14
feat: update validation records

fxrysh:


phenax (changes requested):
A few issues here -
1. You cannot use CNAME together with any other record. It is a limitation with DNS.
2. A record has to be an array so enclose it within square braces.
3. AAAA record type is not supported

Please refer to your hosting service's docs. I couldn't find the proper docs for custom domains with dash but if it wants verification, you'll need to use A and TXT records.

kyuzu:
> A few issues here -
> 
> 1. You cannot use CNAME together with any other record. It is a limitation with DNS.
> 2. A record has to be an array so enclose it within square braces.
> 3. AAAA record type is not supported
> 
> Please refer to your hosting service's docs. I couldn't find the proper docs for custom domains with dash but if it wants verification, you'll need to use A and TXT records.

Sure thing, I'll notify you.

kyuzu:
@phenax

phenax (approved):

PR #1706, opened 2022-02-15
Add security for mattdavison.is-a.dev

ImMattDavison:


phenax (changes requested):
We do not support nested subdomains at the moment. If you need it, you can create a issue for feature request. Can't promise when it will be implemented but till then, this pr will have to be on hold.

TheBotlyNoob:
update: we now allow nested subdomains, but, things like this are on hold until we know it's secure for us to allow these kinds of nested subdomains.

wdhdev:
@ImMattDavison We have just implemented ACME verification records, if you want you can open a new PR adding an ACME verification record and we will merge it ASAP.

PR #1710, opened 2022-02-16
Create chiller.json

Code19Master:


phenax (changes requested):

Also, netlify doesn't support is-a.dev cnames. You can use alternative platforms like railway.app

Code19Master:
Okay, Thanx For the Reply.

domains/chiller.json
```diff
@@ -0,0 +1,10 @@
+{
+  "description": "Website for Discord Bot - Chiller",
+  "owner": {
+    "username": "Code19Master",
+    "email": "divikbabbar@gmail.com"
+  },
+  "record": {
+    "CNAME": "https://chiller.netlify.app"
```
MaskDuck, line 8: 1. netlify doesnt work with is-a.dev, maybe try [railway](https://railway.app)?
2. cname shouldnt be a full url. (cc phenax)

MaskDuck (changes requested):

PR #1725, opened 2022-02-19
Add some more abuse rules

mtgsquad:
This is to let people know, that is-a.dev wasn't made for them to abuse and promote illegal activities with.


MaskDuck:
agreed

phenax (approved):

PR #1742, opened 2022-02-23
Create mudiaga.json

mudiageo:


phenax (changes requested):
Unfortunately, vercel doesn't support is-a.dev subdomain

PR #1743, opened 2022-02-24
Create max.json

NebulaNacho:


phenax (changes requested):
Your domain json is invalid. You can read the documentation and look at existing domain files for reference.

PR #1744, opened 2022-02-24
Create fcircle-doc.json

hiltay:
fcircle-doc


phenax (changes requested):
Also, we need an email address to contact you in case you need to be notified of any changes to is-a-dev. Alternatively, you can keep email as "" and add any of twitter, discord, etc.

hiltay:
thx,I updated it

phenax (approved):

PR #1772, opened 2022-03-01
wbalaile is-a.dev registration

6alaile:
How do I point to a specific subdomain on my website ? All the A records are the same and the subdomain does not have a CNAME record.


mtgsquad:
If you want to point to a subdomain on a pre-existinf site, just use CNAME as shown in our guide, for redirecting to thr subdomain use a URL record.

wilby-mj:
The subdomain doesn't have a CNAME, do I just add one ?

wilby-mj:
Hope this works.

mtgsquad (approved):
Seems good.

PR #1792, opened 2022-03-05
feat: remove kyuzu.json

fxrysh:
I'll be removing my domain since I don't need it anymore. (For now at least)


kyuzu:
On second thought, I still need this domain

PR #1793, opened 2022-03-05
feat: (re)create kyuzu.json

fxrysh:


kyuzu:
I also edit the validation so it's not the same as the old one

phenax (approved):

PR #1794, opened 2022-03-05
Add uiopler.json

uiopler:
Add uiopler.json

### DIFF
--- a/domains/uiopler.json
+++ b/domains/uiopler.json
@@ -1,0 +1,11 @@
+{
+  "description": "uiopler's website",
+  "owner": {
+    "username": "uiopler",
+    "email": "uiopler.dev@gmail.com",
+    "twitter": "uiopler_dev"
+  },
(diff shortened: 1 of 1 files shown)


domains/uiopler.json
```diff
@@ -0,0 +1,11 @@
+{
+  "description": "uiopler's website",
+  "owner": {
+    "username": "uiopler",
+    "email": "uiopler.dev@gmail.com",
+    "twitter": "uiopler_dev"
+  },
+  "record": {
+    "CNAME": "uiopler.netlify.app"
```
TheBotlyNoob, line 9: Netlify doesn't work with a CNAME at the moment, and neither does Vercel. As an alternative, there is [Railway](https://railway.app?referralCode=jj), and [Deta](https://www.deta.sh/)

TheBotlyNoob (changes requested):

PR #1885, opened 2022-03-15
Create epnq.json

wensente27:


AbubakarKang (changes requested):
Even if you don't want to show your email, you will still have to put it under owner but you can leave it as an empty string.

wensente27 (commented):
Thanks

AbubakarKang (approved):
It's good now!

MaskDuck (approved):
O_O your website looks cool

PR #1888, opened 2022-03-16
Update README.md

AbubakarKang:
This is a repository and not a project (although it is, but not for GitHub in this case), so it will be better to change `project` to `repository` in order to make things clearer for generic GitHub users


mtgsquad (approved):
Seems  
Good.

PR #1900, opened 2022-03-17
Badges ✨

s3mant:
better looking readme badges ✨


MaskDuck:
patch 12 wtf

MaskDuck:
patch 69420 when:tm:

MaskDuck:
also can you show an example of the edited file

Bad-Boy-Codes:
> patch 12 wtf

💀

Bad-Boy-Codes:
> patch 12 wtf

wait no
that was my patch
but nvm i created new one

PR #1916, opened 2022-03-18
Create insta.badboy.json

s3mant:
a web app which helps you view insta profile by username


MaskDuck (approved):
uh a bit too sus but acceptable

PR #1935, opened 2022-03-20
Update onichan.json

notdeliable:


domains/onichan.json
```diff
@@ -6,6 +6,6 @@
     "email": "rbstsoul@gmail.com"
   },
   "record": {
-    "CNAME": "hosting.gitbook.io"
+    "CNAME": "onichan.readthedocs.io"
```
MaskDuck, line 9: do you really want to use this cname?

iirc readthedocs cname is `readthedocs.io` for custom domain.

MaskDuck (changes requested):

domains/onichan.json
rbstsoul: let me fix it

rbstsoul: did it

MaskDuck (approved):

PR #1955, opened 2022-03-22
docs: remove `thedev.id`

### LINKED ISSUE #1954: [SUGGESTION] Remove "other services" section
I think the maintainer(s) should remove the "other services" section. This is because `thedev.id` seems unmaintained while `js.org` is only for JS specific project. Maybe we should only keep `runs-on.tech`.

fxrysh:
close #1954


fxrysh:
Should I also remove `js.org` since it's only for JS specific project(s)?

phenax (approved):

phenax:
Thanks!

Issue #489, opened 2021-01-10
[Support] Domain already owned by a vercel account


tejasag:
Your domain name: tejas.is-a.dev

Hi! So I just got my domain name approved and I was trying to configure Vercel to support it, but then I got this error
![image](https://user-images.githubusercontent.com/67542663/104118557-0e757580-5350-11eb-9569-bf36fca14128.png)
Any idea what I can do at this point?

Thanks in advance 
\- Tejas

phenax:
You have a URL redirection record not a subdomain so the link `https://tejas.is-a.dev` will redirect to `https://tejasagarwal.vercel.app` (https://github.com/is-a-dev/register/pull/488/files).
Also, it seems like vercel doesn't support using `is-a.dev` subdomains so you cannot use CNAME record for vercel. https://github.com/is-a-dev/register/issues/249

If you wanted a subdomain (a CNAME record), you will have to use a different hosting service.

tejasag:
Ok thanks! I'll keep the website on vercel for now, but if I find another host, then we can switch to a CNAME record i guess.
But for now, lets keep it as it is. thanks a lot!

phenax:
Alrighty. Closing this issue

anondev-sudo:
I have same problem and I can't understand why does vercel count when someone owns x.domain.com as owner of domain.com

Will someone suggest them to add support for subdomains on different accounts?

TheBotlyNoob:
made (another) discussion about this in https://github.com/vercel/vercel/discussions/6378

TheBotlyNoob:
Reopening because this is kind of annoying.

anondev-sudo:
I tried to use netlify intead of vercel but they are same kind of annoying
![image](https://user-images.githubusercontent.com/85408287/122915582-2d931d00-d35c-11eb-93b3-4bb80000a778.png)

TheBotlyNoob:
NO

Bedrock16:
Hmmm

leecheeyong:
@TheBotlyNoob yea i'm having a kinda same issue but on replit

Bedrock16:
Really I too want it so that I can use it on repl.it

leecheeyong:
I have a solution, which is to notify the user who wants to give up or change the domain to unlink their projects so other users could link their project

IbrahimDevX:
[...]

Issue #249, opened 2020-10-14
Support: Vercel.app custom subdomain issue


tharindusathis:
Can we use this custom domain on Vercel? 
I was trying to add `tharindu.is-a.dev` #188 to a Vercel app which is deployed in http://tharindu.vercel.app, but get the following error, that "domain tharindu.is-a.dev is already owned by another Vercel account.". 
![image](https://user-images.githubusercontent.com/34299693/95962525-19d60500-0e24-11eb-9062-2b1ca3f0a58e.png)

Cannot we use this with Vercel or is there any workaround for it?

phenax:
It seems like they have restrictions for subdomains. Not sure why.

I've found 2 issues there that seem to be related to this - 
* https://github.com/vercel/vercel/issues/3671
* https://github.com/vercel/vercel/issues/3513

But neither of them have a proper resolution.

I'd suggest creating a new issue there.
If they can't help then it does seem like you can't use vercel with is-a-dev subdomains.
Link the issue here so that I can document this problem for anyone else using vercel.

tharindusathis:
https://github.com/vercel/vercel/discussions/5292

phenax:
I've found this other discussion here -> https://github.com/vercel/vercel/discussions/4859
Seems like you could give this a try. Lemme know how it goes.

phenax:
I've found this other discussion here -> https://github.com/vercel/vercel/discussions/4859
Seems like you could give this a try. Lemme know if it helps

tharindusathis:
Reply for the support ticket by Vercel is saying that,

> **The is-a.dev domain is indeed owned by another account on Vercel.**
> **If you believe this be incorrect, can you provide proof of ownership?**

I think that its means may be is-a-dev is using Vercel under that hood? Is it? 

Note: Normally , if someone owns the domain `abc.com` on Vercel lists all the subdomain `xxx.abc.com` as owned by him/her.

phenax:
is-a-dev doesn't use vercel in any way. I think someone else added their `user.is-a.dev` subdomain to vercel. If there is no way around this, you can use is-a-dev to URL redirect to your `.vercel.app` website instead. Not the best way to do it but it doesn't seem like vercel is willing to play ball.

tharindusathis:
Yeah redirecting would be also a fair workaround.  

[...]

PR #2002, opened 2022-03-28
daniel-shuy.json: Use A record instead of CNAME

daniel-shuy:
Hashnode uses Vercel, which prevents `CNAME` from being used (see #249, #489)

If this works, [docs/hosted-at/hashnode.md](https://github.com/is-a-dev/register/blob/main/docs/hosted-at/hashnode.md) should be updated as well

### DIFF
--- a/domains/daniel-shuy.json
+++ b/domains/daniel-shuy.json
@@ -4,6 +4,8 @@
     "email": "daniel_shuy@hotmail.com"
   },
   "record": {
-    "CNAME": "hashnode.network"
+    "A": [
+      "76.76.21.21"
+    ]
   }
 }


MaskDuck (approved):
okay...

(im pretty sure others still can use hashnode then.)

MaskDuck:
also @daniel-shuy would you mind joining our discord server? invite in readme.

i want to talk with you a bit...

daniel-shuy:
Turns out `A` records doesn't work either, looks like `is-a.dev` can't be used with Hashnode since it uses Vercel

@MaskDuck I've joined the Discord

PR #2025, opened 2022-04-01
Update agam778.json

agam778:
Actually, I bought a new domain. So updating cname to redirect to the new domain as I have given this .is-a.dev domain to most places.

Thanks.


MaskDuck (approved):
...

PR #2030, opened 2022-04-01
Update mattw.json

Tweak4141:


MaskDuck (changes requested):
cname cant be used with any other records.

MaskDuck (approved):
n01c3

PR #2043, opened 2022-04-03
Create theanonymous.hacker.json

TheDeveloper7878:


MaskDuck (changes requested):
i had read the index of the file and it is too obvious dangerous.

(for school leaking purpose)

so obviously we can't accept domain such like this.

PR #2050, opened 2022-04-05
Create epicster.json

epicsterrrr:


domains/epicster.json
```diff
@@ -0,0 +1,10 @@
+{
+  "owner": {
+    "username": "epicsterrrr",
+    "email": "",
+    "discord": "Epicster#0001"
+  },
+  "record": {
+    "CNAME": "epicsterrrr.github.io"
```
MaskDuck, line 8: this is already pointed to another custom domain. through, you won't be able to setup your is-a.dev domain.

MaskDuck (changes requested):

PR #2055, opened 2022-04-05
Fix helper CI and other actions improvements.

### LINKED ISSUE #672: [CI] Fix github actions automation for auto-comment and auto-label
When the project started it was using `pull_request_target` event in github actions to trigger some automation. But due to vulnerabilities in the workflow, we are currently using `pull_request` event, because of which the automation stopped working.

The expected flow is

```
PR created ---labeller---> PR labelled (domain) ---merge---> PR comment with instructions for the user
```

andrewstech:
Hello

I have fixed the helper CI and the Labeller CI. I have also added an error message to the checker CI for if it fails.

Have a Good day. 

Did this at 00:46 .

Fixes #672

### DIFF
--- a/.github/workflows/bot.yml
+++ b/.github/workflows/bot.yml
@@ -11,14 +11,28 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Comment
-        if: github.event.action == 'closed' && github.event.pull_request.merged == true
-        uses: actions/github-script@v3
+      - name: Comment PR
+        uses: thollander/actions-comment-pull-request@v1
         with:
-          github-token: ${{secrets.GITHUB_TOKEN}}
-          script: |
-            const { hasLabel } = require(`${process.env.GITHUB_WORKSPACE}/scripts/action-utils.js`);
-            const { instructions } = require(`${process.env.GITHUB_WORKSPACE}/scripts/reply.js`);
-            if (hasLabel(context, 'domain'))
-              console.log('Domain');
-            await instructions(context, github);
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          message: |
+            The changes have been published!! It should reflect in less than 24 hours.
+            ## Here's what you need to do next
+            If your domain points to a server you own, add \`domain-name.is-a.dev\` to your server config. For https, you will have to configure ssl certificate to allow the new subdomain.
+            ### For github pages users,
+            * Go to your github page repo (\`user/user.github.io\`)
+            * Open up the **settings** tab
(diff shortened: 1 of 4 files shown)


andrewstech:
Examples: 
https://github.com/Rehide-Smart/register/pull/3 merged PR with auto message.
https://github.com/Rehide-Smart/register/pull/5 Open PR with error message and labled.

TheBotlyNoob:
This looks okay, but I need to wait for the owner's feedback on this.

cc @phenax

phenax (changes requested):
Thank you for the contribution! Unfortunately this won't work as [the problem](https://github.com/is-a-dev/register/issues/672) that caused us to disable this in the first place would still haunt the changes in the pr. It has to do with how github treats forks for running actions. In this case, we can't access our repos GITHUB_TOKEN from a pull request made by a fork. Let me know if you know of a solution for that issue.

andrewstech:
> Thank you for the contribution! Unfortunately this won't work as [the problem](https://github.com/is-a-dev/register/issues/672) that caused us to disable this in the first place would still haunt the changes in the pr. It has to do with how github treats forks for running actions. In this case, we can't access our repos GITHUB_TOKEN from a pull request made by a fork. Let me know if you know of a solution for that issue.

OK, Some ideas.

For the Helper script we can still use pull_request closed as a trigger. If we remove the line `actions/checkout@v2`
Then the action will  run without any repository/files. So that would stop anyone else from trying to use the github_token.

For the Labeller script we are not checking out any files, so safe there.

For the Checker script Github_token is only used in the error message which is run as a different job also without checkout@v2. 

One other idea would be to make a bot account then use a personal access token from the bot account with only perms to comment, this can then be used instead of GITHUB_TOKEN as a secret.

phenax:
But the problem will still remain because `pull_request` doesn't have access to GITHUB_TOKEN or any other secret.

I've been considering using an external vault-like service to work around this but I have no clue how that would work. It seems like the best bet in our situation. Let me know if you find anything.

Keeping the pr open for the week while we look into this. I'll merge it in after.

andrewstech:
I'm still looking into this.

andrewstech:
Hello me again. @win11bot Is a github account that I created for a different project and is currently being trigged by the checks action. So how is this happing? 

So there is a proxy service running on [heroku](https://is-a-dev.herokuapp.com). The comment action is calling the proxy which hands over a personal access token currently from the @win11bot account. This token has only enabled the repo:public_repos scope, so it can only comment as a normal user with no special privileges. 

More info on the proxy service can be found here: https://github.com/andrewstech/add-pr-comment-proxy .

Let me know your thoughts @phenax .

omxpro:
Wow

PR #2082, opened 2022-04-08
boolean44.json

boolean44-repos:
added boolean44.json


mtgsquad (changes requested):
Please add a valid email, so that we could contact you in the future.

boolean44-repos:
did it fix
if fix = true {
try { accept.PR;
comment "pr accepted"

}}else {
comment "denied"
await.Reply {
if (comment === "ok") {
delete.PR

}

}


}

PR #2107, opened 2022-04-12
Update badboy.json

s3mant:
ok this seems good ig?


MaskDuck (approved):
petition to rename `badboy.json` to `baldboy.json`, react with anything to agree

Bad-Boy-Codes:
> petition to rename `badboy.json` to `baldboy.json`, react with anything to agree

no

PR #2146, opened 2022-04-17
Updated docs and readme

omsenjalia:


VaibhavSys (dismissed):
Gud

omxpro:
Okay!!

omxpro:
I'll wait till it gets merged

hackermondev:
@phenax

phenax (changes requested):
I can see the advantage of having some more information in the readme but this is an unnecassarily long information dump. Don't think this is a good idea as it also makes useful information harder to spot. Alternatively, I think redifining the steps in the readme in a concise manner is the way to go with maybe a couple of examples for more common use-cases.

omxpro:
> I can see the advantage of having some more information in the readme but this is an unnecassarily long information dump. Don't think this is a good idea as it makes useful information harder to spot. Alternatively, I think redifining the steps in the readme in a concise manner is the way to go with maybe a couple of examples for more common use-cases.

It's more useful then the docs folder. Or better just make a docs site

MouseMoosz:
> I can see the advantage of having some more information in the readme but this is an unnecassarily long information dump. Don't think this is a good idea as it makes useful information harder to spot. Alternatively, I think redifining the steps in the readme in a concise manner is the way to go with maybe a couple of examples for more common use-cases.

Adding a TOC should solve that issue.

phenax:
> It's more useful then the docs folder. Or better just make a docs site
Docs site sounds good. Alternatively, github wiki would also make it easier to navigate through the docs but that would be a bit more difficult to contribute to since we can't create prs for wiki changes afaik.
But I'd say currently the docs suffers from bad UX rather than wrong format or platform. Instead of doing all the work required to set up the new platform and then having to maintain it later, we can organize the current docs and create proper indexes to allow users to navigate through them. Even moving your new readme as a single docs file would still hold up quite well (with a ToC as suggested by @MouseMoosz).

> Adding a TOC should solve that issue.
The bigger problem with it being in the README isn't the organization of information but the amount of it. A new user shouldn't be overwhelmed with everything that they need to know just to get started. This is fine however if it is a separate docs file since anyone wanting to go through it is looking for more information.

PR #2170, opened 2022-04-19
Create imd.json

imadboy89:


MaskDuck (changes requested):
wrong directory, put it in `domains/` instead

MaskDuck (approved):

PR #2178, opened 2022-04-20
Create wreeper.json

Wreeper:


domains/wreeper.json
```diff
@@ -0,0 +1,11 @@
+{
+  "description": "wreeper.is-a.dev !!",
+  "repo": "-",
```
MaskDuck, line 3: if you don't have repo, just leave.

MaskDuck (changes requested):

PR #2185, opened 2022-04-21
hi

5rq:
hi


omxpro:
Why there is no repo url?

5rq:
> Why there is no repo url?

good question
I forgot it 🤷‍♂️

domains/syz.json
```diff
@@ -0,0 +1,11 @@
+{
+   "owner":{
+      "username":"Syz",
```
MaskDuck, line 3: your gh username?

domains/syz.json
```diff
@@ -0,0 +1,11 @@
...
+      "discord":"Syz#0128",
+      "email":"Syz@myself.com"
+      "repo":"https://github.com/5rq/5rq.github.io"
```
MaskDuck, line 6: repo should be a dedicated field instead

MaskDuck (changes requested):

omxpro:
F

PR #2191, opened 2022-04-23
remove runs-on.tech

MaskDuck:
this pr remove runs-on.tech, since the project is dead.


TheBotlyNoob:
Do you think you could also add is-a-good-dev?

TheBotlyNoob (approved):

PR #2202, opened 2022-04-24
Create klouqz.json

klouqzz:


RiskyMH:
CNAME doesn't have `http(s)` and any slashes `/`

klouqzcreative:
the site does not work without / i will change it if i can get the domain

PR #2244, opened 2022-05-01
Update william.json

wdhdev:


MaskDuck (changes requested):
we dont support ns record.

PR #2250, opened 2022-05-01
added starnumber profile 

StarNumber12046:


mtgsquad:
A record object is required. If you don't want to use this domain, then you are technically blocking others from using it. Please update it with the required DNS records

MaskDuck (dismissed):
add some records.

mtgsquad (approved):

PR #2253, opened 2022-05-02
Update william.json

wdhdev:


WilliamDavidHarrison:
why is it failing?

MaskDuck (changes requested):
you cant mix cname with other records.

PR #2254, opened 2022-05-02
Create autodiscover.william.json

wdhdev:


MaskDuck (changes requested):
why? we dont support domains for domain validation.

PR #2261, opened 2022-05-02
 Create api.json

a69y:


MaskDuck (changes requested):
can you share more about the project?

icancodebeauty:
Its For My All API's!!, Just Registering Domain For Now!!!

TheBotlyNoob (changes requested):
I don't think we should allow this domain, because `api.is-a.dev` sounds like a legit thing. Maybe try changing the file name to `api.icancodebeauty.json`?

icancodebeauty:
Closing This PR Because Its Legit!!!

PR #2264, opened 2022-05-03
add improvmx email

wdhdev:


MaskDuck (changes requested):
you can't mix cname records with any other records.

WilliamDavidHarrison:
changing to redirect url 1 sec

MaskDuck (approved):

PR #2299, opened 2022-05-06
Add my json file

Imshubh69:


mtgsquad (changes requested):
you can't redirect to a domain that you don't own/doesn't exist.

PR #2303, opened 2022-05-06
Update reefsalty.json

reefsalter:


mtgsquad (changes requested):
You cannot use CNAME and MX together, however, you can simply use nslookup.io to find the ip address for reefsalty.github.io and then point a records to those IP addresses.

mtgsquad (approved):

PR #2306, opened 2022-05-07
sorry, going back to improvmx

boolean44-repos:


boolean44-repos:
can someone merge this

hackermondev (changes requested):
TXT records are not supported. I would recommend just not adding it. ImprovMX doesn't require them iirc.

boolean44-repos:
my last one went through with them

boolean44-repos:
in that case can you close this?

PR #2319, opened 2022-05-08
improvmx

boolean44-repos:


hackermondev:
If you don't mind, I've added a random CNAME there because the validator requires there by a CNAME record or an A record.

boolean44-repos:
ok

hackermondev (approved):

boolean44-repos:
i tried to check but it failed

hackermondev:
> i tried to check but it failed

It will take a few hours to fully update.

boolean44-repos:
it still hasn't gone through

PR #2323, opened 2022-05-09
Create sadix.json

sadix:


MaskDuck (changes requested):
we dont support netlify :(

PR #2338, opened 2022-05-12
Create pinont.host.json

Pinont:


hackermondev (changes requested):
CNAME cannot have ``/`` in it.

PR #2348, opened 2022-05-12
Create Pinont.json

Pinont:


hackermondev (changes requested):
CNAME cannot have ``/`` in it.

PR #2357, opened 2022-05-13
Updated records

demonitize:
Replaced redirect with type A record for glitch because of course i use glitch

### DIFF
--- a/domains/countdown.demonitize.json
+++ b/domains/countdown.demonitize.json
@@ -11,11 +11,6 @@
   "record": {
      "A":[
        "50.31.246.2"
-     ],
-     "MX": [
-      "mx1.improvmx.com",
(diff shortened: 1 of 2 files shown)


hackermondev (changes requested):
TXT records are not supported. You probably don't need them anyways so you can remove them.

demonitize:
I'm using TXT for ImprovMX. also the docs state that you can use TXT records, but only one

demonitize (commented):
removed txt and mx records from countdown subdomain as i already have them on the main domain

demonitize:
@hackermondev @TheBotlyNoob I have submitted the changes that have yet to be reviewed, I would remove the `Pending Changes` tag but I am unable to

PR #2359, opened 2022-05-14
smugthekiler.is-a.dev

SmugTheKiler:


hackermondev (changes requested):
is-a.dev doesn't support vercel. (https://github.com/is-a-dev/register/tree/main/docs#vercel--netlify-note)

PR #2371, opened 2022-05-15
Delete suzumi.badboy.json

s3mant:
i mean , I shifted the website so why take extra domain here.
you guys are already running low 💀
best of luck on future updates.


TheBotlyNoob (approved):
thanks

PR #2396, opened 2022-05-19
Update kekda.json

yxshv:


hackermondev (changes requested):
TXT records are not accepted.

PR #2404, opened 2022-05-21
hi

ItzArman09:


hackermondev (changes requested):
?????? You definitely don't need all these domains.

ItzArman09:
I need

MaskDuck:
you're claiming a bunch of domains... you won't use it anw

MaskDuck:
we still have perms to reject whenever we like

mtgsquad:
@ItzArman09 Claiming a bunch of domains all redirecting to arman.is-a.dev is technically hogging, since one can't have so many projects with such names, and second, you have to respond to our comments, otherwise your previous registration for arman.is-a.dev will also be removed.

ItzArman09:
sorry i had exams so i was not able to reply, also sorry if this is not accepted. and btw can i have all the requests coming to *.arman.is-a.dev redirected to main page?

mtgsquad:
> sorry i had exams so i was not able to reply, also sorry if this is not accepted. and btw can i have all the requests coming to *.arman.is-a.dev redirected to main page?

No, that is not allowed.

ItzArman09:
ah ok thank you. You may close the pr

PR #2432, opened 2022-05-25
Update jeweetzelf.json

jeweetzelfWasTaken:
Added MX and TXT records


MaskDuck (changes requested):
if you want to use any other record, remove cname first.

note that remove cname can affect your site.

jeweetzelfWasTaken:
Hmmm, could I also change the CNAME to an A record (with a numeric ip ofcourse) and then keep the A & MX/TXT records?

MaskDuck:
yes

Issue #2376, opened 2022-05-16
Add requirement: reasonable content


MaskDuck:
theres a lot of users (at least considerable amount of users) that has the page saying "test", "foobar" etc.

there should be a requirements of completed website to grant users subdomain. if not, it would be something called "domain keeping", where you just claim the domain and don't let others use it.

MaskDuck:
this thread will be used to discuss which should qualify for "reasonable content".

phenax:
I agree. But I think it will be pretty difficult to define what this line should be since, pulling up a template to park a sub-domain will not be that difficult. So, at the end it's still up to your judgment to figure out what counts as reasonable. Also, since we've allowed nested sub-domains, the use-cases for these have broadened as well.

The ones that are to be rejected for sure are the single html pages with some temporary text. For the others, I think for the first few non-trivial offenses, we can speak with other maintainers and see if we all agree on whether something should be approved or not. That way, the whole team can develop a consistent judgement for what is to be considered "reasonable".

TheBotlyNoob:
I think that's a good idea

WilliamDavidHarrison:
if someone is obviously just claiming a bunch of subdomains for example: abc, a, b, c then dont merge the prs?

Magniquick:
some ideas -
-  site code must be foss 

- should not contain non-dev related stuff
- should not be "lorem ipsum" / placeholders

MaskDuck:
i dont really agree with the 1st and 2nd one

LogSkele:
yeah, people should be allowed to put non-dev related things

mtgsquad:
What do you mean by "foss"? @Magniquick

MaskDuck:
> What do you mean by "foss"? @ Magniquick

@mtgsquad open sourced

Mosaab-Emam:
I disagree, lets say I am in the process of building my personal website, I would put some placeholder content as a 'placeholder' and update it with my actual code later.
Also, domain keeping is inherently present in whenever domain registering is present, people pay to buy domains they don't immediately use all the time, and the goal clearly is to prevent other people from using it. Honestly I don't know if this should be considered a problem at all.

WilliamDavidHarrison:
This is different, since there are a finite amount of subdomains available, with most people using their name or online name, someone just hogging the domain with nothing/a placeholder on their website is annoying especially if that’s the subdomain they want.

Mosaab-Emam:
[...]

PR #2443, opened 2022-05-27
Create isaiahday.json

Isaiah-Day:
I'm still working on my website (that's why it's kinda sloppy, but i wanted to grab the domain sooner rather than later.

### DIFF
--- a/domains/isaiahday.json
+++ b/domains/isaiahday.json
@@ -1,0 +1,15 @@
+{
+   "description": "Isaiah Day's website",
+  {
+    "owner": {
+      "username": "Isaiah08-D",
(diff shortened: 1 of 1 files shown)


mtgsquad (changes requested):
a. The syntax is wrong
b. You can't reserve a domain for later, as this is considered hogging, stopping someone from using the subdomain, so I'd rather have a proper site before claiming it per  [this](https://github.com/is-a-dev/register/issues/2376#issuecomment-1127533953) + #2376)

Isaiah08-D:
@mtgsquad Ok, i will open this pr again later when I am done.

PR #2456, opened 2022-05-27
Create adricq.json

AdriCQ:


MaskDuck (changes requested):
have some actual content not a placeholder ty

PR #2462, opened 2022-05-28
Create annihilatorrrr.json

annihilatorrrr:


MaskDuck (changes requested):
this is linked to other custom domain already.

remember you can only have 1 custom domain per gh pages.

when you fixed it, lmk!

MaskDuck (approved):
h

PR #2506, opened 2022-06-02
Create higherror.json

HighError:


TheBotlyNoob (approved):
you spelt treasure wrong :)
image

PR #2511, opened 2022-06-02
Create jinso.json

JinsoRaj:


TheBotlyNoob (approved):
:eyes: I like the animations!

PR #2561, opened 2022-06-08
Create sputnick567.json

sputnick567:


hackermondev (approved):

hackermondev (changes requested):
You need to have a valid "email" or "twitter" property in owner. if you have none of the above, or dont wanna share, you can insert others such as discord

PR #2565, opened 2022-06-09
Create anya.json

anyalol:


MouseMoosz:
@anyalol
http://87.98.164.155 doesn't serve anything for me.
`curl "http://87.98.164.155"` also responds with nothing.

PR #2584, opened 2022-06-11
Update alpha.json

YouFoundAlpha:
Added Vercel. I hope it works


TheBotlyNoob (changes requested):
we only support one `TXT` record.

PR #2605, opened 2022-06-15
Add ww and wwww to invalid domains

wdhdev:
These subdomains could be used for abuse if registered


mtgsquad (approved):

MaskDuck (approved):

WilliamDavidHarrison:
@phenax what do you think?

phenax (approved):

PR #2664, opened 2022-06-21
prettier

TheBotlyNoob-ALT:


phenax (changes requested):
We can add prettier just for the script files but let's leave the json files alone. Ideally, the subdomain json files must only have commit logs authored by the owner. I used to run a simple formatter in the initial stages of the project but stopped because of this reason.

PR #2665, opened 2022-06-22
Syz.json

5rq:
I opened a pr a while ago but it had some errors, so I finally fixed them and here is my new pr


domains/syz.json
```diff
@@ -0,0 +1,11 @@
+{
+   "repo":"https://github.com/5rq/5rq.github.io"
+   "owner":{
+      "username":"Syz",
```
MaskDuck, line 4: pls. don't use your nickname. gh username instead. ty.

MaskDuck (changes requested):

PR #2832, opened 2022-07-12
BETA  Add Subdomain:sherikdev

Sherikxd:
Added the domain  sherikdev


andrewstech (changes requested):
Sorry we don't support netlify

PR #2845, opened 2022-07-14
BETA  Add Subdomain:qwkun

Azkun:
Added the domain  qwkun

### DIFF
--- a/domains/qwkun.json
+++ b/domains/qwkun.json
@@ -1,0 +1,11 @@
+
+              {
+                "owner": {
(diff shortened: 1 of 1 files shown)


Azkun:
Uh, I do not understand the way i should register and use is-a.dev, I'd like to know if i made mistakes in the registering and how to correct them. Also i wanna know if the domain name is just gonna be a redirecting domain or not.

MaskDuck (changes requested):

PR #2910, opened 2022-07-19
BETA  Add Subdomain:shubham

deploywithshubham:
Added the domain  shubham


hackermondev (changes requested):
Why is the A record the Cloudfare DNS domain??

PR #3054, opened 2022-08-02
BETA  Add Subdomain:NX

C4NX:
Added the domain  NX


mtgsquad (dismissed):

mtgsquad (changes requested):
Domain Length has to be above 2 chars.

PR #3056, opened 2022-08-02
BETA  Add Subdomain:snowflakeAfk

NyxNoirXD:
Added the domain  snowflakeAfk

### DIFF
--- a/domains/SnowflakeAfk.json
+++ b/domains/SnowflakeAfk.json
@@ -1,0 +1,11 @@
+
+    {
+      "owner": {
+        "username": "Kavindu-Deshapriya",
+        "email": "kavindudeshapriya754@gmail.com"
+      },
+      "record": {
+        "TXT": "2005"
+      }
+    }
+  
--- a/domains/snowflake05.json
+++ b/domains/snowflake05.json
@@ -1,0 +1,12 @@
+
+    {
+      "owner": {
+        "username": "Kavindu-Deshapriya",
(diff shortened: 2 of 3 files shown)


omsenjalia (changes requested):
Invalid TXT records

AceyDoCodes:
> Invalid TXT records 

TXT record can be any string value. You can only have 1 TXT record.
**any string value**
:/ it can be anything
how can it be invalid

omxpro:
> > Invalid TXT records
> 
> `TXT record can be any string value. You can only have 1 TXT record.` **any string value** :/ it can be anything how can it be invalid

Oohh

omsenjalia (dismissed):
TXT records

MaskDuck (changes requested):
can you have some form of actual records instead of some random string?
and also can you elaborate on how will this domain be used/displayed on the online internet? the domains here doesn't seem to work with anything and is just a placeholder (to prevent others to register this domain), which goes against our tos.

also, your file name must be all lowercased. lowercase it.

omxpro:
> can you have some form of actual records instead of some random string? and also can you elaborate on how will this domain be used/displayed on the online internet? the domains here doesn't seem to work with anything and is just a placeholder (to prevent others to register this domain), which goes against our tos.
> 
> also, your file name must be all lowercased. lowercase it.

After you said it he/she still added a TXT record with random string

PR #3093, opened 2022-08-06
Update gxlpes.json

iugx:
Updating the CNAME record.

### DIFF
--- a/domains/gxlpes.json
+++ b/domains/gxlpes.json
@@ -5,7 +5,8 @@
         "email": "guilhermxlopes@gmail.com"
       },
       "record": {
-        "URL": "https://gxlpes.github.io/portfolio-website/"
+        "URL": "https://gxlpes.github.io/portfolio-website/",
+        "CNAME": "gxlpes.github.io"
       }
     }
   


domains/gxlpes.json
```diff
@@ -5,7 +5,8 @@
         "email": "guilhermxlopes@gmail.com"
       },
       "record": {
-        "URL": "https://gxlpes.github.io/portfolio-website/"
+        "URL": "https://gxlpes.github.io/portfolio-website/",
```
VaibhavSys, line 8: You cannot have URL and CNAME record together.
You likely want to use a CNAME record, read more about it [here](https://www.is-a.dev/docs/github_pages).

VaibhavSys (changes requested):

domains/gxlpes.json
gxlpes: Ok, sorry about that.

VaibhavSys: > Ok, sorry about that.
> 
> Do I have to approve your suggestion or something like that?

You have to update the file and fix the record structure, otherwise it cannot be merged. Make sure that the checks are passing.

PR #3106, opened 2022-08-08
Register Subdomain: msg.github.io

neerajinsan804:
Added the domain: msg.github.io


WilliamDavidHarrison:
@neerajinsan804 the domain you want to register will be "msg.github.io.is-a.dev" are you sure that is correct?

PR #3149, opened 2022-08-13
Update instantnoodles.json

NoodlesDev1:
I changed things lol


MaskDuck (changes requested):
if you use cname then pls don't have other record types

PR #3190, opened 2022-08-16
Register Subdomain: Fingann

Fingann:
Added the domain: Fingann


andrewstech (changes requested):
Think you chose the wrong record tipe as A record is ment for IP's. You can use URL for redirect

PR #3201, opened 2022-08-17
Register Subdomain: johnny

john-champagne:
Added the domain: johnny


hackermondev (changes requested):
CNAME should not have `https://` or `/`

PR #3216, opened 2022-08-18
feat: add jonas.is-a.dev

joaomota22:


andrewstech (changes requested):
We don't support vercel due to verification issues

PR #3251, opened 2022-08-23
Register Subdomain: chickenman

chickenmanfy:
Added the domain: chickenman


WilliamDavidHarrison:
@chickenman-is-better-than-you That's not a valid CNAME.

MaskDuck (changes requested):

PR #3306, opened 2022-08-29
Register Subdomain: yagnik

yagnikpt:
Added the domain: yagnik


andrewstech (changes requested):
We don't support vercel due to verification problems

PR #3343, opened 2022-09-01
Register Subdomain: muff

mansoor-roeen:
Added the domain: muff


taha183:
I dont know if A record is correct but i prefer use CNAME if using github pages

MaskDuck (changes requested):
your placeholder website seems to lead to an inaccessible webpage.

lifeless-muffin:
> your placeholder website seems to lead to an inaccessible webpage.

Hey, sorry the site link that it had been hosted on was changed, though now it should be find

mansoor-roeen (commented):
Placeholder site is now accessible

MaskDuck (approved):

PR #3345, opened 2022-09-01
Register Subdomain: sezersinanoglu

sezersinanoglu:
Added the domain: sezersinanoglu

### DIFF
--- a/domains/sezersinanoglu.json
+++ b/domains/sezersinanoglu.json
@@ -6,7 +6,7 @@
       },
 
       "record": {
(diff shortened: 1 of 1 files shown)


taha183:
You have to put your url where you have website. Like you have it on Github pages then use its url. Btw its better to use CNAME or A record with Github pages

LRxDarkDevil:
Beo you are using Github pages then use CNAME it will look like this
reponame.github.io

MaskDuck (approved):

PR #3353, opened 2022-09-02
Register Subdomain: WxC

Emtyfay:
Added the domain: WxC


taha183:
this is totally invalid A record

PR #3357, opened 2022-09-02
Register Subdomain: penieldialu

bilwifi:
Added the domain: penieldialu


LRxDarkDevil:
or i think i just checked something wrong and your CNAME is valid whatever but it doesn't have https and /

PR #3473, opened 2022-09-13
Register Subdomain: borris

linkb15:
Added the domain: borris

### DIFF
--- a/domains/borris.json
+++ b/domains/borris.json
@@ -6,7 +6,7 @@
       },
 
       "record": {
-        "A": ["https://borris.netlify.app/"]
+        "CNAME": ["https://borris.netlify.app/"]
       }
     }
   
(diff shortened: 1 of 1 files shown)


WilliamDavidHarrison:
@linkb15 An A record only has numbers, for example: `8.8.8.8`

WilliamDavidHarrison:
We don't support Netlify sorry.

linkb15:
how about vercel?

WilliamDavidHarrison:
> how about vercel?

No. We do support Railway though.

domains/borris.json
```diff
@@ -0,0 +1,12 @@
+
+    {
+      "owner": {
+        "username": "linkb15",
+        "email": "borris.wiria@gmail.com"
+      },
+
+      "record": {
+        "CNAME": ["https://borris.netlify.app/"]
```
TheBotlyNoob, line 9: We don't support netlify

TheBotlyNoob (changes requested):

PR #3496, opened 2022-09-17
Create boomerangBS.json

boomerangBS:
this is a test


MaskDuck (changes requested):
all i see is "run this repl to see the result here". and it redirect to the replit main page. nah sorry

boomerangBS:
:(

PR #3575, opened 2022-09-30
Register Subdomain: laith

imthatdev:
Added the domain: laith

### DIFF
--- a/domains/laith.json
+++ b/domains/laith.json
@@ -1,0 +1,12 @@
+
+    {
+      "owner": {
+        "username": "babyico",
+        "email": "iconical@programmer.net"
+      },
+
+      "record": {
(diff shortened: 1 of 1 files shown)


domains/laith.json
```diff
@@ -0,0 +1,12 @@
+
+    {
+      "owner": {
+        "username": "babyico",
+        "email": "iconical@programmer.net"
+      },
+
+      "record": {
+        "TXT": "vc-domain-verify=laith.is-a.dev,31bdede79402a2c23012"
```
WilliamDavidHarrison, line 9: Sorry we don't support TXT verification.

wdhdev (changes requested):

TheBotlyNoob (changes requested):
see above.

PR #3627, opened 2022-10-06
Register Subdomain: arnav

arnav-kr:
Added the domain: arnav

### DIFF
--- a/domains/arnav.json
+++ b/domains/arnav.json
@@ -7,7 +7,6 @@
 
       "record": {
         "CNAME": "arnav-kr.github.io",
-        "TXT": "google-site-verification=BIsHFi7TKijw-rS3fXwzHQBATa3PnN0b1b09c0EsdYA"
       }
     }
   
--- a/domains/arnav.json
+++ b/domains/arnav.json
@@ -6,8 +6,8 @@
       },
 
       "record": {
-        "CNAME": ["arnav-kr.github.io"],
-        "TXT": ["google-site-verification=BIsHFi7TKijw-rS3fXwzHQBATa3PnN0b1b09c0EsdYA"]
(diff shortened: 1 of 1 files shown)


wdhdev (changes requested):
404

arnav-kr (commented):
Updated Details

domains/arnav.json
```diff
@@ -6,7 +6,8 @@
       },
 
       "record": {
-        "A": ["199.36.158.100"]
+        "CNAME": ["arnav-kr.github.io"],
```
WilliamDavidHarrison, line 9: CNAME cannot be an array.

domains/arnav.json
```diff
@@ -6,7 +6,8 @@
...
+        "TXT": ["google-site-verification=BIsHFi7TKijw-rS3fXwzHQBATa3PnN0b1b09c0EsdYA"]
```
WilliamDavidHarrison, line 10: Sorry, we do not support TXT verification, also TXT cannot be an array.

wdhdev (changes requested):

domains/arnav.json
arnav-kr: Oh I thought you would support it cause there was option to add a TXT record in the web client so i thought it should be work
IMG_20221007_104112

wdhdev (changes requested):
Sorry, we don't support TXT verification.

arnav-kr (commented):
Done

wdhdev (dismissed):

TheBotlyNoob (dismissed):
:+1:

wdhdev (approved):

MaskDuck (approved):

PR #3639, opened 2022-10-08
Register Subdomain: sterndev

zanemsdev:
Added the domain: sterndev

### DIFF
--- a/domains/sterndev.json
+++ b/domains/sterndev.json
@@ -1,0 +1,12 @@
+
+    {
+      "owner": {
+        "username": "zanemsdev",
+        "email": "r.chevalier.pro@hotmail.com"
+      },
+
+      "record": {
+        "URL": "SterNdev"
+      }
+    }
+  


domains/sterndev.json
```diff
@@ -0,0 +1,12 @@
+
+    {
+      "owner": {
+        "username": "zanemsdev",
+        "email": "r.chevalier.pro@hotmail.com"
+      },
+
+      "record": {
+        "URL": "SterNdev"
```
WilliamDavidHarrison, line 9: Invalid URL

wdhdev (changes requested):

domains/sterndev.json
zanemsdev: URL = https://sterndev.is-a.dev

WilliamDavidHarrison: You cannot use that as a URL as it will create an infinite loop.

PR #3675, opened 2022-10-12
Register Subdomain: patato

PatatoBit:
Added the domain: patato

### DIFF
--- a/domains/patato.json
+++ b/domains/patato.json
@@ -1,0 +1,12 @@
+
+    {
+      "owner": {
+        "username": "PathonScript",
+        "email": "pathonscript@gmail.com"
+      },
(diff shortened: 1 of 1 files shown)


domains/patato.json
```diff
@@ -0,0 +1,12 @@
+
+    {
+      "owner": {
+        "username": "PathonScript",
+        "email": "pathonscript@gmail.com"
+      },
+
+      "record": {
+        "CNAME": "pathonscript.github.io"
```
WilliamDavidHarrison, line 9: You cannot just use a template as your website. Please add some content.

wdhdev (changes requested):

PR #3726, opened 2022-10-18
Register Subdomain: yt-dl

Noma4321:
Added the domain: yt-dl


wdhdev (approved):

andrewstech (changes requested):
We can't merge this due to it violates T&C of another platform.

PR #3730, opened 2022-10-18
Register Subdomain: zendai

n0rs3:
Added the domain: zendai

### DIFF
--- a/domains/zendai.json
+++ b/domains/zendai.json
@@ -1,0 +1,12 @@
+
+    {
+      "owner": {
+        "username": "ZendaiOwl",
+        "email": "victorray91@pm.me"
+      },
+
+      "record": {
+        "CNAME": "zendaiowl.github.io"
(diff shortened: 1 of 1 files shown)


domains/zendai.json
```diff
@@ -0,0 +1,12 @@
+
+    {
+      "owner": {
+        "username": "ZendaiOwl",
+        "email": "victorray91@pm.me"
+      },
+
+      "record": {
+        "CNAME": "zendaiowl.github.io"
```
WilliamDavidHarrison, line 9: 404?

wdhdev (changes requested):

domains/zendai.json
ZendaiOwl: Yes cuz it's setup towards https://website.zendai.ml at the moment so not reachable on the original GH pages URL right now

PR #3746, opened 2022-10-21
Register Subdomain: amr.elrefaiec

amour86:
Added the domain: amr.elrefaiec


wdhdev (changes requested):
404? Also, what do you intend to use on the amr subdomain?

amour86:
I will point them to my VMs on AWS which I'm using to automate some things with the help of Automation Anywhere RPA and these VMs will create some APIs on them to play with it

WilliamDavidHarrison:
> I will point them to my VMs on AWS which I'm using to automate some things with the help of Automation Anywhere RPA and these VMs will create some APIs on them to play with it

We cannot allow you to register the amr subdomain if we don't know specifically what you're doing with them.

hackermondev (approved):

PR #3788, opened 2022-10-26
Register Subdomain: musaubrian

musaubrian:
Added the domain: musaubrian

### DIFF
--- a/domains/musaubrian.json
+++ b/domains/musaubrian.json
@@ -1,0 +1,12 @@
+
+    {
+      "owner": {
+        "username": "musaubrian",
+        "email": "musaubrian45@gmail.com"
+      },
+
+      "record": {
+        "CNAME": "musaubrian.netlify.app"
+      }
+    }
+  


domains/musaubrian.json
```diff
@@ -0,0 +1,12 @@
+
+    {
+      "owner": {
+        "username": "musaubrian",
+        "email": "musaubrian45@gmail.com"
+      },
+
+      "record": {
+        "CNAME": "musaubrian.netlify.app"
```
WilliamDavidHarrison, line 9: Sorry, we don't support Netlify, but you can jave your domain redirect to your Netlify site using a `URL` record.

wdhdev (changes requested):

domains/musaubrian.json
musaubrian: Thanks for the feedback

hackermondev (changes requested):
Netlify is not supported.

PR #3800, opened 2022-10-27
Register Subdomain: richard

Jets0m:
Added the domain: richard

### DIFF
--- a/domains/richard.json
+++ b/domains/richard.json
@@ -1,0 +1,12 @@
+
+    {
+      "owner": {
+        "username": "Jets0m",
+        "email": "rpbutler@gmail.com"
+      },
+
+      "record": {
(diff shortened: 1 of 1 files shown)


domains/richard.json
```diff
@@ -0,0 +1,12 @@
+
+    {
+      "owner": {
+        "username": "Jets0m",
+        "email": "rpbutler@gmail.com"
+      },
+
+      "record": {
+        "CNAME": "richardbutler.netlify.app"
```
WilliamDavidHarrison, line 9: Sorry, we don't support netlify.

wdhdev (changes requested):

hackermondev (changes requested):
Netlify is not supported

PR #3804, opened 2022-10-28
Register Subdomain: *.queirozt

QueirozT:
Added the domain: *.queirozt


wdhdev (changes requested):
Sorry, we don't support adding wildcard domains.

PR #3825, opened 2022-10-30
Remove my domain (`emma.json`)

triphora:
Originally added in #650. My previous motivation for this domain is not relevant anymore because I now have https://emmacypress.com going to the same content. I would also like to open this domain up to anyone else who wants it, as I've never actually used this.


wdhdev (approved):
Sorry to see you go

andrewstech (approved):

PR #3959, opened 2022-11-16
prettier

wdhdev:


WilliamDavidHarrison:
@phenax What do you think?

(I spent like an hour doing this since prettier was being stupid lol)

TheBotlyNoob (dismissed):
Quoting @phenax (https://github.com/is-a-dev/register/pull/2664#pullrequestreview-1048149693):
> We can add prettier just for the script files but let's leave the json files alone. Ideally, the subdomain json files must only have commit logs authored by the owner. I used to run a simple formatter in the initial stages of the project but stopped because of this reason.

WilliamDavidHarrison:
It doesn't include JSON files

TheBotlyNoob:
ah, I didn't see the file diff, sorry.

TheBotlyNoob (dismissed):

WilliamDavidHarrison:
> ah, I didn't see the file diff, sorry.

Np

WilliamDavidHarrison:
Oh, we need Phenax to review as well

TheBotlyNoob:
yeah...

WilliamDavidHarrison:
@phenax

WilliamDavidHarrison:
It's because I was modifying script files, I think.

scripts/certbot.sh
```diff
@@ -1,6 +1,5 @@
-#!/usr/bin/env bash
```
phenax, line 1: Shebang needs to be here

phenax (changes requested):
Please add a `format` command in package.json scripts so that we have one consistent method for formatting

WilliamDavidHarrison:
> Please add a `format` command in package.json scripts so that we have one consistent method for formatting

I was using prettier online, not a desktop version. Also, prettier isn't installed in `package.json`.

scripts/verify-record.sh
```diff
@@ -1,5 +1,3 @@
-#!/usr/bin/env sh
```
phenax, line 1: Needs the shbang

phenax (changes requested):
This and there's a merge conflict in README

PR #4064, opened 2022-11-27
Register Subdomain: sameerahmed

sxmeer-ahmed:
Added the domain: sameerahmed

### DIFF
--- a/domains/sameerahmed.json
+++ b/domains/sameerahmed.json
@@ -1,0 +1,11 @@
+ 
+    {
+      "owner": {
+        "username": "sxmeer-ahmed",
+        "email": "ahmedkhansameer50@gmail.com"
+      },
+      "record": {
+        "TXT": "786"
+      }
+    }
+      


domains/sameerahmed.json
```diff
@@ -0,0 +1,11 @@
+ 
+    {
+      "owner": {
+        "username": "sxmeer-ahmed",
+        "email": "ahmedkhansameer50@gmail.com"
+      },
+      "record": {
+        "TXT": "786"
```
WilliamDavidHarrison, line 8: Please provide a valid record

wdhdev (changes requested):

domains/sameerahmed.json
VaibhavSys: > Please provide a valid record

It is a valid record.

VaibhavSys (approved):

domains/sameerahmed.json
WilliamDavidHarrison: > > Please provide a valid record
> 
> It is a valid record.

It isn't as it is not used for anything and it's their only record.

VaibhavSys: We don't know what he could be using it for, its valid as in DNS and he could be doing something of his own with that record.

WilliamDavidHarrison: > We don't know what he could be using it for, its valid as in DNS and he could be doing something of his own with that record.

We have to know what he's using it for, it's literally in the docs.

WilliamDavidHarrison:
No changes that meet our requirements

PR #4283, opened 2022-12-27
Register Subdomain: mrbeast

LegendryTiger:
Added the domain: mrbeast

### DIFF
--- a/domains/mrbeast.json
+++ b/domains/mrbeast.json
@@ -1,0 +1,11 @@
+ 
+    {
+      "owner": {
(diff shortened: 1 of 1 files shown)


MaskDuck (changes requested):
1. why are you registering a name of a well-known youtuber? are you afraid of being sued?
2. the site which you're leading to is a NSFW website. please change that.

i'm too nice to give you a chance to fix tbh instead of denying and closing PR

LegendryTiger:
Ong

PR #4285, opened 2022-12-27
Register Subdomain: solo

Hadimhz:
Added the domain: solo


wdhdev (changes requested):
You cannot use a local IP

Hadimhz:
oops i was messing with the website didn't know it actually create a pr

PR #4356, opened 2023-01-02
Register Subdomain: _github-pages-challenge-lucifer1708.sumit

lucifer1708:
Added the domain: _github-pages-challenge-lucifer1708.sumit

### DIFF
--- a/domains/_github-pages-challenge-lucifer1708.sumit.json
+++ b/domains/_github-pages-challenge-lucifer1708.sumit.json
@@ -1,0 +1,11 @@
+ 
+    {
+      "owner": {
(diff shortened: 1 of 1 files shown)


omsenjalia (changes requested):
I do not think this is allowed

TerraPlayz:
> > I do not think this is allowed
> 
> it's allowed

Alright 👍

omsenjalia (approved):
👍👍

wdhdev (changes requested):
sumit.is-a.dev isn't registered so you cannot register a verification record for it.

Please register sumit.is-a.dev first.

PR #4404, opened 2023-01-05
Made 2 commits to change the URL TO CNAME to prevent redirection to another website

ALBINPRAVEEN:
## Requirements
Unless explicitly specified otherwise by a **maintainer**, your domain must pass **all** of the indicated requirements below.

- [ ] The file is in the `domains` folder and is in the JSON format.
- [ ] You have completed your website. 
- [ ] The website is reachable.  
- [ ] You're not using Vercel or Netlify.
- [ ] There is sufficient information at the `owner` field.

### DIFF
--- a/domains/albinpraveen.json
+++ b/domains/albinpraveen.json
@@ -5,7 +5,7 @@
         "email": "albinpraveen135790@gmail.com"
       },
       "record": {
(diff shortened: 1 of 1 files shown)


wdhdev (changes requested):
Sorry we don't support Vercel so the only way to use it is as a redirect.

PR #4406, opened 2023-01-05
Updated A record on hatkidchan.is-a.dev

hatkidchan:
## Requirements
Unless explicitly specified otherwise by a **maintainer**, your domain must pass **all** of the indicated requirements below.

## Link to Website

https://hatkidchan.is-a.dev/

### DIFF
--- a/domains/hatkidchan.json
+++ b/domains/hatkidchan.json
@@ -7,7 +7,6 @@
     "discord": "hkc#0517"
   },
   "record": {
-    "A": ["5.181.108.112"],
-    "TXT": "aHR0cHM6Ly9oYXRraWRjaGFuLmlzLWEuZGV2L21lb3ctMGUwYWZkOWYtMzdlOC00ODc4LWJmYWYtOGE2MjAyYTI5ODFl"
+    "A": ["5.181.108.112"]
   }
 }
--- a/domains/hatkidchan.json
+++ b/domains/hatkidchan.json
@@ -1,5 +1,5 @@
 {
-  "description": "Static personal page of hatkidchan (hkc) (thx Mia)",
+  "description": "kc personal website",
   "owner": {
     "username": "hatkidchan",
     "email": "hatkidchan@gmail.com",
@@ -7,7 +7,7 @@
     "discord": "hkc#0517"
   },
   "record": {
-    "A": ["216.128.176.102"],
+    "A": ["5.181.108.112"],
     "TXT": "aHR0cHM6Ly9oYXRraWRjaGFuLmlzLWEuZGV2L21lb3ctMGUwYWZkOWYtMzdlOC00ODc4LWJmYWYtOGE2MjAyYTI5ODFl"
   }
 }


omsenjalia (commented):
What is the usage of A and TXT records??

TerraPlayz:
What are you using them for?

domains/hatkidchan.json
```diff
@@ -1,13 +1,13 @@
...
+  "description": "kc personal website",
   "owner": {
     "username": "hatkidchan",
     "email": "hatkidchan@gmail.com",
     "twitter": "hatkidchan",
     "discord": "hkc#0517"
   },
   "record": {
-    "A": ["216.128.176.102"],
+    "A": ["5.181.108.112"],
```
WilliamDavidHarrison, line 10: I can't reach this IP

wdhdev (changes requested):
We don't support TXT verification and please read comment below.

TerraPlayz:
> We don't support TXT verification and please read comment below.

I don't think it's TXT verification. It's just TXT record.

TerraPlayz:
> > What is the usage of A and TXT records??
> 
> > A
> 
> It's, well, address of the server, bruh
> 
> > TXT
> 
> may be small ARG, or may be not. left it for future use

For security reasons we don't allow TXT verification as William says so remove it until you decide you want to add that in future.

domains/hatkidchan.json
WilliamDavidHarrison: @hatkidchan I need to know what the IP is for.

hatkidchan: It is IP of my server. No idea why you can't reach it, but it works for everyone else

image

TXT record is *not* used for verification

hatkidchan: > It's for a server. It's not a webserver

It *is* a webserver and I already configured it a while ago
image

WilliamDavidHarrison: Ok. But what is the TXT for?

hatkidchan: > Ok. But what is the TXT for?

as already said

> may be small ARG, or may be not. left it for future use

According to [docs](https://www.is-a.dev/docs/domain_structure/) I can put anything I want here, so it's just base64-encoded url

TerraPlayz: Yes it is a Base64 encoded URL

MaskDuck (dismissed):
i'm perfectly fine with A records, but can you explain me how is the TXT records being used? we need to have an actual reason for every record added in our service. 

note that docs are not final, maintainer discretion is final.

domains/hatkidchan.json
TerraPlayz: > @WilliamDavidHarrison @TerraPlayz i see perfectly **NOTHING** wrong with A records as it can be used for various purposes.

I know MaskDuck i did know that as i have previously worked with them as i am a OS developer and i have used A ip's before

hatkidchan:
TXT record is used as a small easter egg and nothing else. If it's *that* critical, sure, I'll remove it now. It's not added just now, it was added with previous commit and there was no questions why and what it is

hatkidchan:
> It just redirects to small hello

Thanks captain, and I see no problem in that. Documentation *clearly* states that it can be `any string value`: 
image

If using TXT records in some way is prohibited and/or requires detailed explanation of every bit in it, maybe it also should be stated in docs?

omsenjalia (approved):
Issue solved

wdhdev (approved):

PR #4520, opened 2023-01-13
Register Subdomain: firework

ghost:
Added the domain: firework


Firewoork:
i accidently added another thinking it was broken but firefox disabled popups

PR #4530, opened 2023-01-14
Register Subdomain: toxicrecker

toxicrecker:
Added the domain: toxicrecker


wdhdev (commented):
We don't support Vercel or domain verification.

PR #4728, opened 2023-02-06
Added lucky via is-a-dev-cli

imluckydev:
Added lucky via the CLI


wdhdev (changes requested):
Please use a valid email.

PR #4764, opened 2023-02-11
Register monkey.is-a.dev

hbt272:
Added `monkey` via the registration website.


wdhdev (changes requested):
Please use a valid record

PR #4777, opened 2023-02-12
Register ihzan.is-a.dev

rwixy:
Added `ihzan` via the registration website.


wdhdev (changes requested):

Ihsan-N:
i already added cname also my domai is ihzan not ihsan404 please check it

wdhdev (approved):

PR #4794, opened 2023-02-13
Register firas.is-a.dev

fraxnet:
Added `firas` via the registration website.


wdhdev (changes requested):
Please use a valid record

PR #4801, opened 2023-02-14
Register ksetrar.is-a.dev

Ksetrar:
Added `ksetrar` via the registration website.


wdhdev (changes requested):
Please use a valid A record

PR #4818, opened 2023-02-15
Register spotify.is-a.dev

whoisjayd:
Added `spotify` via the registration website.


wdhdev (changes requested):
Sorry you cannot register a domain with the name of a registered trademark.

PR #4841, opened 2023-02-16
Register yoli.is-a.dev

JollyJolli:
Added `yoli` via the registration website.


wdhdev (changes requested):
Internal server error

PR #4858, opened 2023-02-18
Register nikit.is-a.dev

nikitsoni:
Added `nikit` via the registration website.


wdhdev (changes requested):
Please use a valid record

PR #4894, opened 2023-02-20
Register ellie.is-a.dev

junepark678:
Added `ellie` via the registration website.


wdhdev (changes requested):
I can't reach your website

PR #4897, opened 2023-02-21
Register all-day-hi.is-a.dev

intfsq:
Added `all-day-hi` via the registration website.


wdhdev (changes requested):
Please use a valid URL

PR #4900, opened 2023-02-21
suki.json : Update NS Records

SukiPlayz:


wdhdev (changes requested):
NS records are not MX records, we don't support NS either.

PR #4910, opened 2023-02-22
Register yatochka.is-a.dev

yatochka-dev:
Added `yatochka` via the registration website.


wdhdev (changes requested):
Sorry we don't support Vercel.

PR #4922, opened 2023-02-24
Register verydrstrange.is-a.dev

MinecraftOS:
Added `verydrstrange` via the registration website.


wdhdev (changes requested):
Please use a different URL

PR #4953, opened 2023-02-26
Register mehdi.is-a.dev

essmehdi:
Added `mehdi` via the registration website.


wdhdev (changes requested):
We don't support Vercel or TXT verification

PR #4954, opened 2023-02-26
Added trip via is-a-dev-cli

ov-ah:
Added trip via the CLI


wdhdev (changes requested):
We do not support NS records.

PR #4955, opened 2023-02-26
Register techno.is-a.dev

RealTechno:
Added `techno` via the registration website.


wdhdev (changes requested):
Please use valid records

PR #4964, opened 2023-02-27
Register juang.is-a.dev

JuanG970:
Added `juang` via the registration website.


wdhdev (changes requested):
Please use valid records

PR #4978, opened 2023-02-28
Register sxrv.is-a.dev

savar95x:
Added `sxrv` via the registration website.


wdhdev (changes requested):
Please use a valid record

wdhdev (changes requested):
Please use a valid record for sxrv.is-a.dev

wdhdev (approved):

PR #4990, opened 2023-03-01
Register neuthral.is-a.dev

neuthral:
Added `neuthral` via the registration website.


wdhdev (changes requested):
You can't just have a template as your website

PR #5095, opened 2023-03-18
Create pixyshare.json

emmu123:
All done

Requesting SSL secure

### DIFF
--- a/domains/pixyshare.json
+++ b/domains/pixyshare.json
@@ -1,0 +1,11 @@
+ 
+    {
+      "owner": {
+        "username": "emmu",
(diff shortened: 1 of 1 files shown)


wdhdev (changes requested):
What is this website used for?

emmu123:
> What is this website used for?

used for ordering codes

emmu123:
This website not abuse because this site not using any login or register method. This site asking some Data if can give only they need to use.
Data means not high value data's
Only Asking Discord user name, tag and user id.

wdhdev (approved):

PR #5157, opened 2023-03-28
Create indianprime.json

emmu123:
All done

This site not using illegal

### DIFF
--- a/domains/indianprime.json
+++ b/domains/indianprime.json
@@ -1,0 +1,10 @@
+{
+
+  "owner": {
+    "username": "emmu",
+    "email": "aneenababu71901@gmail.com"
+  },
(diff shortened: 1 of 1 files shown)


wdhdev (changes requested):
Sorry, we cannot allow you to register a domain as hosting services can abuse our domains and also, things like your panel URL are not even up.

emmu123:
> Sorry, we cannot allow you to register a domain as hosting services can abuse our domains and also, things like your panel URL are not even up.

not abuse this was info about hosting

emmu123:
> Sorry, we cannot allow you to register a domain as hosting services can abuse our domains and also, things like your panel URL are not even up.

and i not connect panel .

PR #5227, opened 2023-04-06
Register anay.is-a.dev

anay-208:
Added `anay.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


Anay028:
I closed this pull due to the reason that I use React.js and I m hosting on vercel which I realized after seeing the docs after creating the pull request

PR #5244, opened 2023-04-08
Register meowice.is-a.dev

MeowIce:
Added `meowice.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


wdhdev (changes requested):
Please use a valid record

PR #5280, opened 2023-04-14
Register volcanicdev.is-a.dev

ghost:
Added `volcanicdev.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


wdhdev (changes requested):
Please use a valid record

PR #5287, opened 2023-04-15
Register verify.is-a.dev

nagdongdev-0713:
Added `verify.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).

### DIFF
--- a/domains/verify.json
+++ b/domains/verify.json
@@ -1,0 +1,10 @@
+{
+    "owner": {
+        "username": "nagdongdev-0713",
+        "email": "lyw8459@gmail.com"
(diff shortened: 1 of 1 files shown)


wdhdev (changes requested):
This site looks like a scam website, so we will not be allowing this domain to be registered.

nagdongdev-0713:
This site is a Discord Authorization (Role Assignment) site. Use hCaptcha.

WilliamDavidHarrison:
The website may lead users into believing it is an official is-a.dev site.

PR #5336, opened 2023-04-19
Register joaodeveloper.is-a.dev

JoaoDeveloper1:
Added `joaodeveloper.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


wdhdev (changes requested):
1) You cannot delete another user's domain.
2) I cannot reach your website, it resolves in a 404.

PR #5345, opened 2023-04-20
Register drax.is-a.dev

m1kuwu:
Added `drax.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


wdhdev (changes requested):
Please use a valid record

PR #5438, opened 2023-04-26
Register merlinisadev.is-a.dev

merlincod:
Added `merlinisadev.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


wdhdev (changes requested):
Please use a valid record.

PR #5529, opened 2023-05-01
Register casperong.is-a.dev

CasperOng:
Added `casperong.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


wdhdev (changes requested):

PR #5598, opened 2023-05-12
Register aayush.is-a.dev

Aayush117:
Added `aayush.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


wdhdev (changes requested):
Sorry, aayush.is-a.dev is taken.

PR #5605, opened 2023-05-14
Register comphack.is-a.dev

COMP-HACK:
Added `comphack.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


wdhdev (changes requested):
Please provide valid a record.

PR #5673, opened 2023-05-25
Register owner.is-a.dev

raspberry-hash:
Added `owner.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


wdhdev (changes requested):
I cannot allow you to register this domain due to abuse risks.

PR #5680, opened 2023-05-26
Update jeydin.json

jeydinpham:
I want to make my link redirect to my new link as people will undoubtedly still use this link


Jeydin21:
I see it failed let me try something else

PR #5690, opened 2023-05-26
Register domathid.is-a.dev

DomathID:
Added `domathid.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


mtgsquad (changes requested):
Registering, on incorrect domain,delete in the file with is-a.dev, and add other records.

DomathID:
Okay

PR #5696, opened 2023-05-27
Register domathdotid.is-a.dev

DomathID:
Added `domathdotid.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


WilliamDavidHarrison (changes requested):

WilliamDavidHarrison (approved):

PR #5758, opened 2023-06-01
Register rick.is-a.dev

EnderPoint07:
Added rick.is-a.dev using the Discord bot.


wdhdev (changes requested):
1. You can't just have a rick roll redirect.
2. You need to register `astley.is-a.dev` first.

EnderPoint07:
> 1. You can't just have a rick roll redirect.
so im not allowed to do this? even if i register astley.isa.dev?

WilliamDavidHarrison:
You’re not allowed to just have a rick roll redirect.

PR #5800, opened 2023-06-04
Switch to Cloudflare Pages and add Google Search Console

HeroGamers:
## Requirements

## Website Link/Preview

https://portfolio-4cw.pages.dev/

### DIFF
--- a/domains/marcus.json
+++ b/domains/marcus.json
@@ -1,12 +1,13 @@
 {
     "description": "A portfolio website for HeroGamers (Marcus).",
-    "repo": "https://github.com/HeroGamers/HeroGamers",
+    "repo": "https://github.com/HeroGamers/Portfolio",
     "owner": {
         "username": "HeroGamers",
         "email": "webmaster@herogamers.dev",
         "twitter": "ItsHeroGamers"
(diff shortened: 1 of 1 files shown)


HeroGamers:
I don't know why the test doesn't pass, it mentions that the CNAME record should be a hostname, but it is. The stacktrace seems to show it's something else causing the issue.

wdhdev (approved):
I've removed the TXT record as we don't support TXT verification.

HeroGamers:
Ah okay, is that a technical thing with the TXT record not being supported, or due to not wanting the verification records on?

I actually added it because I was looking through others on the repo to see if Cloudflare Pages would work, and saw that others had the Google search console record on it.

TXT is listed as a valid property on the docs though. But either way, at the end of the day it doesn't really matter to me, was just curious.

WilliamDavidHarrison:
We don't support TXT verification due to sometimes with TXT verification it verifies the entire is-a.dev domain, which would mean other users can't use that service. We do support TXT records for other things like for email, etc.

Also, CNAME records can't mix with TXT records due to DNS requirements.

PR #5822, opened 2023-06-07
Register collin.is-a.dev

lxmmel:
Added `collin.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).

### DIFF
--- a/domains/sunshine.json
+++ b/domains/sunshine.json
@@ -5,6 +5,6 @@
     },
 
     "record": {
-        "A": ["192.168.0.101"]
+        "URL": "https://hanyaku.glitch.me"
     }
 }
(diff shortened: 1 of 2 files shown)


domains/sunshine.json
```diff
@@ -0,0 +1,10 @@
+{
+    "owner": {
+        "username": "Hanyaku-Chan",
+        "email": "collinbuchkamer@gmail.com"
+    },
+
+    "record": {
+        "A": ["192.168.0.101"]
```
WilliamDavidHarrison, line 8: You cannot use a private IP address.

WilliamDavidHarrison, line 8: You cannot use a private IP address.

wdhdev (changes requested):

domains/sunshine.json
Hanyaku-Chan: My bad

wdhdev (approved):

PR #5852, opened 2023-06-11
Added Email Forwarding Using ImproveMX

Dheirya:
Added email forwards from my domain to my email using the ImproveMX site; let me know if there's anything wrong and I need to change/fix!

### DIFF
--- a/domains/dheirya.json
+++ b/domains/dheirya.json
@@ -7,6 +7,8 @@
         "twitter": "DheiryaTyagi"
     },
     "record": {
-        "CNAME": "dheirya.pages.dev"
+        "CNAME": "dheirya.pages.dev",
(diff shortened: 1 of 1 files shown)


GoodBoyNeon:
You cannot have both CNAME and MX in one domain

wdhdev (changes requested):
You can't use CNAME and other records together.

Dheirya:
Oh ok so I can't have both a website and forwarding email?

WilliamDavidHarrison:
You can, just not with a CNAME.

Dheirya:
Oh so I have to use "URL" in record? Doesn't that just redirect that to the original site? If so I don't want to add email forwarding from this domain...

PR #5863, opened 2023-06-12
Register ayaanhaider.is-a.dev

ayaanhaider-dev:
Added `ayaanhaider.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


wdhdev (changes requested):

PR #5935, opened 2023-06-20
Register status.is-a-dev.is-a.dev

status-dev:
Added `status.is-a-dev.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


wdhdev (changes requested):
This domain is already in use.

PR #5946, opened 2023-06-22
Register .is-a.dev

status-dev:
Added `.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


wdhdev (changes requested):
You can’t have CNAME and TXT records together.

PR #6128, opened 2023-07-10
BETA: Register swoth.is-a.dev

Swothh:
Added `swoth.is-a.dev` using the Discord bot.


wdhdev (changes requested):
We don't support Vercel.

PR #6129, opened 2023-07-10
Register quickcut.is-a.dev.is-a.dev

Queaxtra:
Added `quickcut.is-a.dev.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


wdhdev (changes requested):
We don't support Vercel.

PR #6201, opened 2023-07-15
Register kio.is-a.dev

YoruAkio:
Added `kio.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).

### DIFF
--- a/domains/kio.json
+++ b/domains/kio.json
@@ -1,0 +1,10 @@
+{
+    "owner": {
+        "username": "YoruAkio",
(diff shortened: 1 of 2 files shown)


wdhdev (changes requested):
We don't support Vercel

YoruAkio (commented):
can we use two domain with one CNAME, is that even allowed?

wdhdev (approved):

PR #6202, opened 2023-07-15
Register tts.deva.is-a.dev.is-a.dev

turbomaster95:
Added `tts.deva.is-a.dev.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


turbomaster95:
only mer

PR #6204, opened 2023-07-15
Register ryzm.is-a.dev

wvvb:
Added `ryzm.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


hackermondev (changes requested):
You need a CNAME/A/URL record

PR #6223, opened 2023-07-16
BETA: Register techlewi.is-a.dev

LewiDev:
Added `techlewi.is-a.dev` using the site.


wdhdev (changes requested):
Please use a valid record. Our [docs](https://www.is-a.dev/docs) can help you.

PR #6228, opened 2023-07-16
BETA: Register l7neg.is-a.dev

L7NEG:
Added `l7neg.is-a.dev` using the site.

### DIFF
--- a/domains/l7neg.json
+++ b/domains/l7neg.json
@@ -1,0 +1,10 @@
+{
+        "owner": {
+           "username": "L7NEG",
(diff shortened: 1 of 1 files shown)


wdhdev (changes requested):
We don't support Vercel.

L7NEG:
Yes I see but I want to suggest to make a dns website with automatic api

WilliamDavidHarrison:
We don't support Vercel on our domains unfortunately.

L7NEG:
Ok np, thanks for the reply.

PR #6233, opened 2023-07-17
BETA: Register m-dashboard.is-a.dev

mobin-mazahery-school:
Added `m-dashboard.is-a.dev` using the site.


wdhdev (changes requested):
Please use valid records.

PR #6243, opened 2023-07-17
BETA: Register legion.is-a.dev

LeGi0N09:
Added `legion.is-a.dev` using the Discord bot.


wdhdev (changes requested):
We don't support Vercel.

PR #6257, opened 2023-07-17
BETA: Update aleph.is-a.dev

ghost:
Updated `aleph.is-a.dev` using the site.


alephdfilms:
fixed with A records since CNAME fucking sucks dude ("Unfortunately, due to DNS restrictions CNAME records with other records do not work. But, you can reverse lookup the CNAME and use the A records it points to instead. So then you can use other records.")

alephdfilms:
fucking hell it failed

PR #6275, opened 2023-07-18
BETA: Register hyun.is-a.dev

hyun790:
Added `hyun.is-a.dev` using the site.


VaibhavSys (changes requested):
Parking a domain for no reason isn't allowed.

PR #6301, opened 2023-07-19
BETA: Update mako.is-a.dev

Kauntar:
Updated `mako.is-a.dev` using the site.


wdhdev (changes requested):
A records need to be in arrays.

PR #6393, opened 2023-07-23
BETA: Register akshay.is-a.dev

ak4pr:
Added `akshay.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
This domain is already taken and we do not support Vercel either.

PR #6396, opened 2023-07-23
BETA: Register azuria.is-a.dev

xvea:
Added `azuria.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
We do not support Vercel.

PR #6416, opened 2023-07-24
BETA: Register yashrao.is-a.dev

YashRao105:
Added `yashrao.is-a.dev` using the [dashboard](https://manage.is-a.dev).


mtgsquad (changes requested):
The whole filename must be lowercase, please keep that in mind when registering a domain. cc @andrewstech We need to build this into the manage.is-a.dev website.

wdhdev (changes requested):
I cannot reach your website.

PR #6433, opened 2023-07-24
BETA: Register arenawars.is-a.dev

arenawars2023:
Added `arenawars.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/arenawars.json
+++ b/domains/arenawars.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "arenawars2023",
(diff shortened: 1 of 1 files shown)


WilliamDavidHarrison (changes requested):
I'm unsure if we can allow your website due to it wanting to download a exe file.

arenawars2023:
> I'm unsure if we can allow your website due to it wanting to download a exe file.

hmmm

PR #6481, opened 2023-07-27
Register play-prozt.is-a.dev

pxldevv:
Added `play-prozt.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).

### DIFF
--- a/domains/play-prozt.json
+++ b/domains/play-prozt.json
@@ -1,0 +1,10 @@
+{
+    "owner": {
+        "username": "Techpro709",
+        "email": "devzt709@gmail.com"
+    },
+
+    "record": {
+        "TXT": "minehut-verification=a4edd8e6d8945bfe9082cf093ca70ade56974917ee1747e288676d914808001b"
+    }
+}


wdhdev (changes requested):

Techpro709:
why not, isn't it just a txt record though?

WilliamDavidHarrison:
It verifies that you own the domain which is why we cannot allow it.

Techpro709:
I see, the way this works (correct me if I am wrong) is that it checks that you own the SUB domain. Honestly even if it checks for the whole domain, i don't know why you would care anyways. I can only register subdomains by submitting PR. I just need minehut (minecraft hosting) to know that I can forward people from the domain to the minecraft server. 

Honestly that is how most places work for domains. If this does go against your TOS thats fine, I want to find out if there is a way to work around this, or make an exception.

WilliamDavidHarrison:
Unfortunately most TXT verification systems then verify the whole domain so we cannot allow it at all.

Techpro709:
Do you think you could check if minehut does this. I really would like this domain for this minecraft server.

WilliamDavidHarrison:
I cannot check that as I do not personally use MineHut.

PR #6487, opened 2023-07-27
BETA: Register utils.is-a.dev

Blob2763:
Added `utils.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
Please use a valid record.

wdhdev (approved):

PR #6531, opened 2023-07-31
BETA: Update sematary.is-a.dev

h3artcrushz:
Updated `sematary.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):

PR #6537, opened 2023-07-31
BETA: Register rahad.is-a.dev

gazirahad7:
Added `rahad.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
Please use valid records. We also don't support Vercel.

PR #6568, opened 2023-08-01
BETA: Update viniciusmanoel.is-a.dev

vncsmnl:
Updated `viniciusmanoel.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
We do not support Vercel

PR #6614, opened 2023-08-04
BETA: Register just-muzz.is-a.dev

just-Muzz:
Added `just-muzz.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
Please use valid record.

PR #6615, opened 2023-08-04
BETA: Update mbg.is-a.dev

MBG1337:
Updated `mbg.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):

PR #6698, opened 2023-08-05
Register krinzal.is-a.dev

krinzal:
Added `krinzal.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


wdhdev (approved):
I have switched your record to a `URL` record because we do not support Vercel as a CNAME.

PR #6718, opened 2023-08-05
BETA: Update i2rys.is-a.dev

I2rys:
Updated `i2rys.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
We don't support Vercel.

PR #6785, opened 2023-08-05
BETA: Register creeperplayer20.is-a.dev

creeperplayer20:
Added `creeperplayer20.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
You cannot use a private IP address.

Issue #6627, opened 2023-08-04
Support : The domain has been verified by another user/organization


fdciabdul:
Your domain name: fdciabdul.is-a.dev

Explain any issues you are facing

github sent me like this 

![image](https://github.com/is-a-dev/register/assets/31664438/708d3864-a72c-4f5e-aba5-06cf9fca075d)

even my domain is still in there 

![image](https://github.com/is-a-dev/register/assets/31664438/e5b4121b-9460-4e94-bda9-1728d95ac674)

then when i trying to verify the CNAME ,this happen

![image](https://github.com/is-a-dev/register/assets/31664438/a8306537-1913-4e07-b7b2-d1fd3bdb8609)

i don't know

vikramv20:
![Screenshot_20230805-030001](https://github.com/is-a-dev/register/assets/80647839/3383bf12-f717-4772-8625-9a5e2589014a)

I got this problem too.

Jesgran:
got this problem too

NatBLida68:
same issue got the mail just now

ItachiSan:
Same here!

Adolar0042:
just got the mail too

techpixel:
Got the email as well

seriaati:
same issue.

antton-dev:
Same problem for me

ejaz4:
+1 here, just got an email with the exact same thing.
![image](https://github.com/is-a-dev/register/assets/73548455/77dc3562-2cd6-495f-8918-d997a023b35d)

loiSvelasco:
Got the same problem too :<

vncsmnl:
Got the same problem too here !!!

zyrouge:
Same issue here too.

WilliamDavidHarrison:
Hi everyone, we are currently looking into this. We do not know why this has happened.

boredcodebyk:
got the same mail

404Dev-404:
got this email too i had a mini heart attack when i saw it

AldenizenMC:
same issue

ShivangKakkar:
Same!

Shinyzenith:
Recieved this mail too. Says I have 7 days to resolve the issue before someone else can claim it.

WilliamDavidHarrison:
I have found the issue that caused it, PR #6624. I have reverted the pull request, but I believe everyone will need to re-verify their domains with GitHub Pages.

[...]

PR #6797, opened 2023-08-06
Update rv178.json

rv178:
#6627


wdhdev (changes requested):
https://github.com/is-a-dev/register/issues/6627#issuecomment-1666382002

rv178:
It is in lowercase though. Also, that is the correct format for the filename right?

wdhdev (approved):

WilliamDavidHarrison:
> It is in lowercase though. Also, that is the correct format for the filename right?

Sorry, I think I requested changes on the wrong PR, my bad.

PR #6815, opened 2023-08-06
BETA: Register nyde.is-a.dev

justnyde:
Added `nyde.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
We don't support Vercel

PR #6859, opened 2023-08-07
BETA: Register google.is-a.dev

ayushbite:
Added `google.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
I cannot allow you to register `google.is-a.dev` as Google is a registered company and trademark.

PR #6908, opened 2023-08-08
BETA: Register izzahxfast.is-a.dev

Ryzaa07:
Added `izzahxfast.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
Please use valid records.

wdhdev (changes requested):
Please use valid records for all of your domains.

PR #6910, opened 2023-08-08
BETA: Register test.is-a.dev

jojo-995:
Added `test.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
Please use a valid record and use a different domain name.

PR #6923, opened 2023-08-08
BETA: Register scratch-coins.is-a.dev

ghost:
Added `scratch-coins.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
I cannot allow this domain to be registered due to abuse risks.

PR #6939, opened 2023-08-08
BETA: Update sgm-uptime.is-a.dev

ShadowGaming100:
Updated `sgm-uptime.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):

wdhdev (changes requested):

PR #6941, opened 2023-08-08
Create bn2wq5qkv26d.tnt.json - tnt blog p2

kernel010:
## Requirements

- [ x] The file is in the `domains` folder and is in the JSON format.
- [ x] You have completed your website. 
- [ x] The website is reachable.  
- [ x] You're not using Vercel or Netlify.  
- [ x] The CNAME record doesn't contain `https://` or `/`.  
- [ x] There is sufficient information at the `owner` field.

### DIFF
--- a/domains/bn2wq5qkv26d.tnt.json
+++ b/domains/bn2wq5qkv26d.tnt.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "SebTNT",
+        "email": "turquoisetntmultimedia@gmail.com"
+    },
+    "record": {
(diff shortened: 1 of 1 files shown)


wdhdev (changes requested):
We don't support domain verification.

SebTNT:
Sorry just so i can understand for future reference, how come verification is not supported? I would’ve thought it still works as it uses a CNAME to a google domain - so would it not be treated as any other CNAME?

WilliamDavidHarrison:
In some cases, it can actually verify the entire is-a.dev domain to that account, so that's why we cannot allow it.

PR #6963, opened 2023-08-09
Register se.is-a.dev

Bobbbyyyyy:
Added `se.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


wdhdev (changes requested):
1. Please use a valid record
2. I cannot allow this domain to be registered as it uses Discord OAuth.

PR #6970, opened 2023-08-09
BETA: Register ramadan.is-a.dev

Ramadan-Hassan:
Added `ramadan.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
Please use a valid record.

PR #7064, opened 2023-08-13
BETA: Register china.is-a.dev

kerryhuang2013:
Added `china.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
Please use valid records.

PR #7156, opened 2023-08-16
update json file

SerStars:


wdhdev (changes requested):
You will need to use Verify by HTTPS for verifying your domain with Discord.

PR #7172, opened 2023-08-17
BETA: Register snowyjs.is-a.dev

clarkpy:
Added `snowyjs.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
Please use a valid record.

PR #7173, opened 2023-08-17
BETA: Register _discord.stefdp.is-a.dev

Stef-DP:
Added `_discord.stefdp.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
Sorry, we don't support TXT verification. You will need to verify by HTTPS.

PR #7177, opened 2023-08-17
BETA: Register _discord.window5.is-a.dev

JSDforJ:
Added `_discord.window5.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
We don't support Discord verification, you will need to use verify by HTTPS.

PR #7211, opened 2023-08-18
BETA: Register kuru.emre.is-a.dev

kuru-emre:
Added `kuru.emre.is-a.dev` using the [dashboard](https://manage.is-a.dev).


domains/kuru.emre.json
WilliamDavidHarrison, line 1: You will need to register `emre.is-a.dev` first.

wdhdev (changes requested):

PR #7218, opened 2023-08-19
BETA: Update iammithani.is-a.dev

MuhammadUsmanMithani:
Updated `iammithani.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
1. Please use a valid record, which you had before
2. For Discord verification you will need to use verify by HTTPS.

PR #7238, opened 2023-08-19
BETA: Register iftach.is-a.dev

iftachhh:
Added `iftach.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
We don't support Vercel.

PR #7240, opened 2023-08-20
Update imat.json

d3vadam:


wdhdev (changes requested):
We don't support CNAME records with other records.

wdhdev (changes requested):
I can't reach your website

wdhdev (approved):

PR #7282, opened 2023-08-22
BETA: Register exocubeyt.is-a.dev

ExoCubeYT:
Added `exocubeyt.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/x6rh3pe4jkng.exocubeyt.json
+++ b/domains/x6rh3pe4jkng.exocubeyt.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "ExoCubeYT",
+        "email": "umanbhat7780@gmail.com"
+    },
+    "record": {
+        "CNAME": "gv-rynyaxcxi6w6lk.dv.googlehosted.com"
+    }
+}
--- a/domains/exocubeyt.json
+++ b/domains/exocubeyt.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "ExoCubeYT",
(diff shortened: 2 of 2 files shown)


domains/x6rh3pe4jkng.exocubeyt.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "ExoCubeYT",
+        "email": "umanbhat7780@gmail.com"
+    },
+    "record": {
+        "CNAME": "gv-rynyaxcxi6w6lk.dv.googlehosted.com"
```
WilliamDavidHarrison, line 7: We don't support TXT verification

wdhdev (changes requested):

domains/x6rh3pe4jkng.exocubeyt.json
ExoCubeYT: But blogger needs it so what do I do, also blogger says this

We have not been able to verify your authority to this domain. On your domain registrar's website, locate your Domain Name System (DNS) settings and enter the following two CNAMEs: (Name: exocubeyt, Destination: ghs.google.com) and (Name: x6rh3pe4jkng, Destination: gv-rynyaxcxi6w6lk.dv.googlehosted.com). See https://support.google.com/blogger/answer/1233387 for detailed instructions. Support page link

WilliamDavidHarrison: We can't allow verification records unfortunately, so you will need to use another service.

PR #7299, opened 2023-08-23
BETA: Register jermaine.is-a.dev

Codemaine:
Added `jermaine.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
We don't support Vercel.

PR #7367, opened 2023-08-26
BETA: Register ds+gamer.is-a.dev

ds-gamer:
Added `ds+gamer.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
You cannot have a `+` in your domain name.

PR #7368, opened 2023-08-26
BETA: Register ilxvmorkq245.stefdp.is-a.dev

Stef-DP:
Added `ilxvmorkq245.stefdp.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
We don't support domain verification.

PR #7378, opened 2023-08-27
BETA: Register simplysnox .is-a.dev

SimplySnox:
Added `simplysnox .is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):

PR #7379, opened 2023-08-27
BETA: Register phantombot.is-a.dev

ghost:
Added `phantombot.is-a.dev` using the Discord bot.


wdhdev (changes requested):

PR #7430, opened 2023-08-28
BETA: Update soldierp.is-a.dev

HippoDev03:
Updated `soldierp.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
Please fix your records manually as I think there is a bug with the dashboard. Please also note, you cannot combine a CNAME record with other records.

PR #7472, opened 2023-08-29
BETA: Register jesulonimii.is-a.dev

jesulonimii:
Added `jesulonimii.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):

PR #7515, opened 2023-08-31
Update hifzur.is-a.dev

hifzurmax:
Updated `hifzur.is-a.dev` using the [dashboard](https://manage.is-a.dev).


luni-moon (changes requested):
There is no page here, just a GitHub Pages 404, we don't reserve domains.

wdhdev (changes requested):
I cannot reach your website.

PR #7516, opened 2023-08-31
Update flyingcheese41.json

FlyingCheese41:
Change TO TXT RECORD BECAUSE WEB IS NOT WORKING IN DOMAIN


wdhdev (changes requested):

PR #7531, opened 2023-09-01
Create _discord.oxmc.json

oxmc:
## Requirements

## Website Link/Preview

This is for the connections part in discord

### DIFF
--- a/domains/_discord.oxmc.json
+++ b/domains/_discord.oxmc.json
@@ -1,0 +1,10 @@
+{
+    "owner": {
+        "username": "oxmc",
+        "email": "oxmc7769.mail@gmail.com",
+        "discord": "oxmc7769"
+    },
+    "record": {
+        "TXT": "dh=6493e80ff0645ec8c706643ae6e357066becb660"
+    }
+}


oxmc:
@WilliamDavidHarrison When running this checks, i'm a bit confused, we can make the files like this correct? or is that a thing yet.

wdhdev (changes requested):
We don't support domain verification, you will need to use the Verify by HTTPS option instead.

WilliamDavidHarrison:
> @WilliamDavidHarrison When running this checks, i'm a bit confused, we can make the files like this correct? or is that a thing yet.

We don't support Discord TXT verification at this time unfortunately. I will pass this onto @phenax and we might be able to get this added.

oxmc:
Thank you for informing me!

oxmc:
> We don't support domain verification, you will need to use the Verify by HTTPS option instead.

it does not seem to like that

image

WilliamDavidHarrison:
Does the file end in `.html`? It shouldn't. Also, try giving it some time as that has fixed it for other users as well.

oxmc:
Still nothing:
image

WilliamDavidHarrison:
Unfortunately I cannot provide support with this as this is not to do with your subdomain, it is to do with GitHub Pages and Discord.

PR #7562, opened 2023-09-02
BETA: Update peacy.is-a.dev

Peaqcy:
Updated `peacy.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
It seems you are trying to verify your domain with Discord. Unfortunately we do not support that as of right now, and you will need to use the verify by HTTPS option Discord provides instead.

wdhdev (approved):

PR #7569, opened 2023-09-02
BETA: Register void6670.is-a.dev

void6670:
Added `void6670.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
We cannot allow this website as it goes against YouTube's ToS.

PR #7651, opened 2023-09-06
BETA: Register mertcelik.is-a.dev

sw3do:
Added `mertcelik.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
We do not support Vercel or domain verification.

PR #7792, opened 2023-09-12
Register samuelsiv.is-a.dev

samuelsiv:
Added `samuelsiv.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


VaibhavSys (changes requested):
That's a local IP Address

PR #7811, opened 2023-09-13
CI: URL Validation

VaibhavSys:
This pull request adds a workflow which will run weekly on every Monday and can be ran manually too. The workflow will go through all the domains and if that domain is a redirect (contains a URL redirect record), it will check if that URL is reachable or broken. It produces a few false positives sometimes. It is multi threaded and insanely fast. As of now, it can validate the whole repository in just ~4.5 minutes.

It seperates the result into 3 parts: valid urls, non valid urls, and non-http ulrls. (Supports both HTTP and HTTPS). For each url, its file path is also mentioned.

### DIFF
--- a/domains/phuc.json
+++ b/domains/phuc.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "phucbm",
+        "email": "phucbm.dev@gmail.com"
+    },
+    "record": {
+        "URL": "https://buiminhphuc.com"
+    }
+}


wdhdev (dismissed):
Looks good to merge, @phenax

VaibhavSys:
@phenax Please review this when you are free

tests/url-validation.py
```diff
@@ -0,0 +1,57 @@
...
+                urls_data["invalid"][file_path] = url
+                print(f"URL '{url}' in file '{file_path}' is not reachable.")
+        else:
+            urls_data["non-http"][file_path] = url
+            print(f"URL '{url}' in file '{file_path}' is neither HTTP nor HTTPS.")
+
+
+for root, _, files in os.walk(domains_directory):
+    for filename in files:
+        thread = threading.Thread(target=handle_url_validation, args=(root, filename))
```
phenax, line 50: Creating 1000s of physical threads is a terrible idea even for a one-shot script. Can we instead use some sort of batching that processes things maybe 10-20 at a time?

VaibhavSys: #8145

PR #7822, opened 2023-09-13
BETA: Register lucious.is-a.dev

eumw:
Added `lucious.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):

PR #7885, opened 2023-09-16
BETA: Register ovidiu-radu.is-a.dev

Oxidiu:
Added `ovidiu-radu.is-a.dev` using the [dashboard](https://manage.is-a.dev).


DEV-DIBSTER (dismissed):

wdhdev (changes requested):
We do not support NS records.

PR #7907, opened 2023-09-17
feat(script): adds support for aaaa records

### LINKED ISSUE #7617: Support AAAA records
I just enabled IPv6 on my AWS EC2 instance and opened a pull request (#7614) to add the IPv6, but noticed it failed a check - it appears AAAA records aren't supported.

We need more websites to support IPv6 if we want to move away from IPv4, and I believe the effects of IPv4 address exhaustion are already known. I won't explain the benefits of IPv6 again because there's entire articles, indeed, entire websites, dedicated to doing such.

Thanks~

phenax:
Adds support for AAAA records

Fixes #7617


VaibhavSys (approved):
@andrewstech CLI and Dashboard will need to be updated accordingly

phenax:
@is-a-dev/maintainers, this introduces support for `AAAA` records in addition to having a slightly more strict check for `A` records. For the most part, `AAAA` are going to be the same as `A` with the only difference being that it would be an array of ipv6 addresses instead of ipv4.

We should keep an eye on any new `AAAA` records being introduced to see if they play along with everything else.

PR #7925, opened 2023-09-17
BETA: Register om-srivastava.is-a.dev

Ilikepizza2:
Added `om-srivastava.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
Please use a valid record.

PR #7954, opened 2023-09-18
BETA: Register cdn-stefdp.is-a.dev

Stef-DP:
Added `cdn-stefdp.is-a.dev` using the [dashboard](https://manage.is-a.dev).


Stef-00012:
what's the issue that causes the checks to fail?

PR #7957, opened 2023-09-18
Update bday-list.suki.json

SukiPlayz:
Added a TXT record with a CNAME


wdhdev (changes requested):

PR #7993, opened 2023-09-20
BETA: Register felix-the-relix.is-a.dev

soliwoli:
Added `felix-the-relix.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
We don't support Vercel.

PR #8004, opened 2023-09-20
Register sushi.is-a.dev

sushi-ae:
Added `sushi.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).

### DIFF
--- a/domains/sushi.json
+++ b/domains/sushi.json
@@ -1,0 +1,10 @@
+{
+    "owner": {
+        "username": "sushi-ae",
+        "email": "frapdotpng@gmail.com"
+    },
+
+    "record": {
+        "CNAME": "sushi-ae.github.io"
+    }
+}
--- a/domains/sushi.json
+++ b/domains/sushi.json
@@ -5,6 +5,6 @@
     },
 
     "record": {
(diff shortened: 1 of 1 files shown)


domains/sushi.json
```diff
@@ -0,0 +1,10 @@
+{
+    "owner": {
+        "username": "sushi-ae",
+        "email": "frapdotpng@gmail.com"
+    },
+
+    "record": {
+        "CNAME": "sushi-ae.github.io"
```
DEV-DIBSTER, line 8: Just redirects to our site. Please have an actual site.

DEV-DIBSTER (changes requested):

domains/sushi.json
sushi-ae: why does it do that, I didn't even set an alternate domain yet

sushi-ae: even if I unpublish the site, sushi-ae.github.io still redirects to is-a.dev

sushi-ae: any ideas why this happening? or how to potentially fix it?

PR #8025, opened 2023-09-20
BETA: Register sonycovery.is-a.dev

sonycovery:
Added `sonycovery.is-a.dev` using the [dashboard](https://manage.is-a.dev).


domains/pro5rb.json
VaibhavSys, line 1: Empty file

VaibhavSys (changes requested):

PR #8100, opened 2023-09-23
BETA: Register danix.is-a.dev

ghost:
Added `danix.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/danix.json
+++ b/domains/danix.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "danix421",
+        "email": "danix@disroot.org"
+    },
+    "record": {
+        "CNAME": "danix421.github.io"
+    }
+}


danix421:
already used is-a.dev in the past (before the new dashboard was made), though I removed my domain a few months back since I wasn't using it a lot.

domains/danix.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "danix421",
+        "email": "danix@disroot.org"
+    },
+    "record": {
+        "CNAME": "danix421.github.io"
```
DEV-DIBSTER, line 7: 404 CNAME record.

DEV-DIBSTER (changes requested):

domains/danix.json
danix421: it's meant to be like that. I haven't commited the page itself yet, since I wanted to work around the domain first.

WilliamDavidHarrison: @danix421 We do not support domain reservation, so you will need to at least put something there like maybe some basic contact details and info before you can get the domain.

PR #8145, opened 2023-09-23
CI: URL Validation

VaibhavSys:
Continuation of #7811
This pull request adds a workflow which will run weekly on every Monday and can be ran manually too. The workflow will go through all the domains and if that domain is a redirect (contains a URL redirect record), it will check if that URL is reachable or broken. It produces a few false positives sometimes. It is multi threaded and uses batching capped at 20 workers and is insanely fast. As of now, it can validate the whole repository in just ~3.0 minutes. It will produce an artifact which will be retained for 6 days.

It seperates the result into 3 parts: valid urls, non valid urls, and non-http ulrls. (Supports both HTTP and HTTPS). For each url, its file path is also mentioned.


.github/workflows/url-validation.yml
```diff
@@ -0,0 +1,26 @@
...
+      with:
+        python-version: '3.10'
+    - run: pip install requests
+    - run: python tests/url-validation.py
+    - name: Upload Artifact
+      uses: actions/upload-artifact@v3
+      with:
+        name: URL Validiation Result
+        path: url-validation-result.json
+        retention-days: 6
```
WilliamDavidHarrison, line 26: I think it would be best to keep the uploaded artifact for a full 7 days.
```suggestion
        retention-days: 7
```

wdhdev (changes requested):

.github/workflows/url-validation.yml
VaibhavSys: I realized that after making the PR and going to sleep 😅

wdhdev (approved):

phenax (approved):
Looks great!

PR #8403, opened 2023-10-02
BETA: Register π.is-a.dev

4pii4:
Added `π.is-a.dev` using the Discord bot.


wdhdev (changes requested):
We do not support special characters in the file name. Also your file name must be 2 characters or longer.

PR #8727, opened 2023-10-16
BETA: Register test.is-a.dev

wany-oh:
Added `test.is-a.dev` using the [dashboard](https://manage.is-a.dev).


domains/test.json
DEV-DIBSTER, line 1: Will not be releasing this domain.

DEV-DIBSTER (changes requested):

PR #8764, opened 2023-10-17
BETA: Register that.is-a.dev

lsy1998:
Added `that.is-a.dev` using the [dashboard](https://manage.is-a.dev).


domains/that.json
DEV-DIBSTER, line 1: Domain is generic, will not be releasing.

DEV-DIBSTER (changes requested):

PR #8819, opened 2023-10-19
Register bimarsha,is-a.dev.is-a.dev

Bimarshaaa:
Added `bimarsha,is-a.dev.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


DEV-DIBSTER (changes requested):
Invalid A record, you placed a host name instead of IPv4.

PR #8821, opened 2023-10-19
BETA: Register admin.denboy0123.is-a.dev

juqwtf:
Added `admin.denboy0123.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
Please use valid records.

PR #8851, opened 2023-10-20
BETA: Register server.is-a.dev

LemonZest686:
Added `server.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
Please use a valid record and choose a different subdomain due to abuse risks.

PR #8885, opened 2023-10-22
Register equinox.is-a.dev

Chibigon:
Added `equinox.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


VaibhavSys (changes requested):
We don't support netlify

PR #8912, opened 2023-10-23
BETA: Register ev.is-a.dev

imdopee:
Added `ev.is-a.dev` using the [dashboard](https://manage.is-a.dev).


VaibhavSys (changes requested):
We don't allow hostings.

PR #8925, opened 2023-10-23
Update owner field for all of my domains

Stef-DP:
switched the mail to my custom mail `admin@stefdp.is-a.dev`, remove `"twitter"` field where present and added a `"discord"` field with my discord ID

if possible could you delete all history commits where is visible the main gmail that i used on all of those domains before now?


Stef-00012:
if possible could you delete all history commits where is visible the main gmail that i used on all of those domains before now?

wdhdev (approved):

WilliamDavidHarrison:
> if possible could you delete all history commits where is visible the main gmail that i used on all of those domains before now?

Unfortunately this is not possible.

PR #8927, opened 2023-10-23
BETA: Register valentino.is-a.dev

abcdefghijklmnopqrstuval:
Added `valentino.is-a.dev` using the [dashboard](https://manage.is-a.dev).


VaibhavSys (changes requested):
We don't support Vercel Domain Verification

PR #8960, opened 2023-10-26
BETA: Register nexon.is-a.dev

vn-nexonDev:
Added `nexon.is-a.dev` using the [dashboard](https://manage.is-a.dev).


VaibhavSys (changes requested):
Invalid A Records. It must be an array. Refer to the docs

PR #8687, opened 2023-10-14
BETA: Register deadcode.is-a.dev

RichardKanshen:
Added `deadcode.is-a.dev` using the [dashboard](https://manage.is-a.dev).

PR #8989, opened 2023-10-27
BETA: Register zb92427382.deadcode.is-a.dev

RichardKanshen:
Added `zb92427382.deadcode.is-a.dev` using the [dashboard](https://manage.is-a.dev).
As a side note, this is a request for domain verification for an email address connected to [an already approved subdomain `deadcode`](https://github.com/is-a-dev/register/pull/8687). I believe email addresses are allowed here, since MX records can be requested here too, so I hope I did not do anything wrong :P

### DIFF
--- a/discord.deadcode.json
+++ b/discord.deadcode.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "RichardKanshen",
+        "email": "richard@kanshen.click"
+    },
+    "record": {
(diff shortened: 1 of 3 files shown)


RichardKanshen:
As a side note, this is a request for domain verification for an email address. I believe email addresses are allowed here, since MX records can be requested here too, so I hope I did not do something wrong :P

PR #8999, opened 2023-10-27
BETA: Register datasets.is-a.dev

pranavagrawal321:
Added `datasets.is-a.dev` using the [dashboard](https://manage.is-a.dev).


VaibhavSys (changes requested):
Distributing pirated content is not allowed.

PR #9025, opened 2023-10-28
BETA: Register theking.is-a.dev

CC52-dev:
Added `theking.is-a.dev` using the [dashboard](https://manage.is-a.dev).


VaibhavSys (changes requested):

CC52-dev:
> Reserving domains is not allowed

I am not reserving domains
I have an ip address to an web server
I want to put that in a is-dev domain

PR #9050, opened 2023-10-30
BETA: Register sqlmerr.is-a.dev

sqlmerr:
Added `sqlmerr.is-a.dev` using the [dashboard](https://manage.is-a.dev).


VaibhavSys (changes requested):
You cannot create an infinite redirect loop.

PR #9051, opened 2023-10-30
BETA: Register conycord.is-a.dev

mehvmt:
Added `conycord.is-a.dev` using the [dashboard](https://manage.is-a.dev).


VaibhavSys (changes requested):
We do not allow these types of domains

PR #9061, opened 2023-10-30
BETA: Register dunkiegaming.is-a.dev

DunkieGaming:
Added `dunkiegaming.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
We do not allow hosting services websites.

PR #9077, opened 2023-10-31
Update sx9.json

satr14washere:
too lazy to fill in the promt just for a cname change so heres a description

cant access website? short.io is a url redirect service too use a custom domain to redirect to another site

im changing from the URL record since its not working

### DIFF
--- a/domains/sx9.json
+++ b/domains/sx9.json
@@ -5,6 +5,6 @@
         "email": "sx-91@outlook.com"
     },
     "record": {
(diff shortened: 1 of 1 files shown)


VaibhavSys:
Your domain doesn't seem to be in the [DNS](https://dnschecker.org/all-dns-records-of-domain.php?query=sx9.is-a.dev&rtype=ALL&dns=google). Thats weird

mtgsquad (approved):

PR #9130, opened 2023-11-03
Update deadcode.is-a.dev - Added ImprovMX records

RichardKanshen:
Added ImprovMX mx and SPF (TXT) records.
Moved CNAME to a `www.` sub-record.

### DIFF
--- a/domains/www.deadcode.json
+++ b/domains/www.deadcode.json
@@ -4,7 +4,6 @@
         "email": "richard@kanshen.click"
     },
     "record": {
-        "MX": ["mx1.improvmx.com", "mx2.improvmx.com"],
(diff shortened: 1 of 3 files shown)


domains/www.deadcode.json
WilliamDavidHarrison, line 1: We do not support domains starting with `www` unfortunately.

wdhdev (changes requested):

domains/www.deadcode.json
RichardKanshen: Moved the CNAME back to deadcode.json, and MX to mail.deadcode.json, does that look alright?

wdhdev (approved):

PR #9142, opened 2023-11-04
BETA: Register h1r0.is-a.dev

XxHiroTTVxX:
Added `h1r0.is-a.dev` using the [dashboard](https://manage.is-a.dev).


VaibhavSys (changes requested):

PR #9145, opened 2023-11-04
BETA: Register medmh.is-a.dev

medragneel:
Added `medmh.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/medmh.json
+++ b/domains/medmh.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "medragneel",
+        "email": "medmystgon@gmail.com"
+    },
+    "record": {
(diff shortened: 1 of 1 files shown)


domains/medmh.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "medragneel",
+        "email": "medmystgon@gmail.com"
+    },
+    "record": {
+        "CNAME": "netlify.app"
```
DEV-DIBSTER, line 7: We do not support netlify. Also that is there normal site.

DEV-DIBSTER (changes requested):

PR #9234, opened 2023-11-07
BETA: Register rupandhungana.is-a.dev

rupandhungana:
Added `rupandhungana.is-a.dev` using the [dashboard](https://manage.is-a.dev).


andrewstech (changes requested):
Sorry we don't support Vercel.

PR #9239, opened 2023-11-07
Update lucaledd.json

display5:
## Requirements

### DIFF
--- a/domains/lucaledd.json
+++ b/domains/lucaledd.json
@@ -4,8 +4,6 @@
         "email": "scorpinyesyeson@gmail.com"
     },
     "record": {
-        "A": ["217.174.245.249"],
-        "MX": ["hosts.is-a.dev"],
-        "TXT": "v=spf1 a mx ip4:217.174.245.249 ~all"
+        "CNAME": "lucaforever.is-a.dev"
     }
 }


domains/lucaledd.json
```diff
@@ -4,8 +4,6 @@
         "email": "scorpinyesyeson@gmail.com"
     },
     "record": {
-        "A": ["217.174.245.249"],
-        "MX": ["hosts.is-a.dev"],
-        "TXT": "v=spf1 a mx ip4:217.174.245.249 ~all"
+        "CNAME": "lucaforever.is-a.dev"
```
DEV-DIBSTER, line 7: Service just showcases another domain on the same service?

DEV-DIBSTER (changes requested):

domains/lucaledd.json
andrewstech: The other domain also uses hosting by is-a.dev however as the server will detect the domain that is being used as this one the content won't mirror both domains they will still be separate

lucaforever: tried registering both domains as they are both my aliases, but didn't want to make a separate server for each of them

PR #9515, opened 2023-11-21
Register vpn.is-a.dev

chalotech:
Added `vpn.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


wdhdev (changes requested):

PR #9619, opened 2023-11-25
Create _acme-challenge.kleb.is-a.dev.json

KrazyKleb:
## This is supposed to be a CNAME record! The website says so


DEV-DIBSTER:
@andrewstech @WilliamDavidHarrison Could you look into this one?

WilliamDavidHarrison:
I don't think we will be accepting this domain because it seems to be a CNAME DNS zone sort of thing.

PR #9621, opened 2023-11-25
Create saturn

Zer0Rea1:


NotRana:
which mistake i am doing

VaibhavSys (changes requested):
The filename should end with `.json`. Rename the file to `saturn.json` inside domains

PR #9684, opened 2023-11-28
BETA: Register vpn.lylythechosenone.is-a.dev

lylythechosenone:
Added `vpn.lylythechosenone.is-a.dev` using the [dashboard](https://manage.is-a.dev).


lylythechosenone:
Just have to say, I'm not exactly sure what I'm doing here, so the record may be completely wrong. This is simply meant to be a domain for a WireGuard VPN server.

lylythechosenone:
Actually, scratch this. I've found another option, and I'll use that rather than waste your resources here.

PR #9756, opened 2023-12-01
Update ppy.is-a.dev

ringo360:
Updated `ppy.is-a.dev` using the [dashboard](https://manage.is-a.dev).


ringo360:
*It looks like CNAME could not handle the domain, so we decided to use hosting.
(Sorry bad English)

DEV-DIBSTER (approved):

PR #9847, opened 2023-12-04
Create astra.json

LilyMusic:
domain


domains/astra.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "LilyMusic",
+        "discord": "astraop"
+    },
+    "record": {
+        "AAAA": ["167.86.83.49:6210"]
```
VaibhavSys, line 7:
```suggestion
        "A": ["167.86.83.49"]
```

VaibhavSys (changes requested):
Ports cannot be added in A or AAAA records and AAAA are for IPv6 not IPv4

domains/astra.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "LilyMusic",
+        "discord": "astraop"
+    },
+    "record": {
+        "CNAME": ["proxy.private.danbot.host"]
```
DEV-DIBSTER, line 7: CNAME should be only a string, no array.

DEV-DIBSTER (changes requested):

PR #9981, opened 2023-12-10
BETA: Register settings.is-a.dev

hooktodiscord:
Added `settings.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/panel.json
+++ b/domains/panel.json
@@ -1,0 +1,12 @@
+{
+    "owner": {
+        "username": "hooktodiscord",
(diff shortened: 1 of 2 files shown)


domains/panel.json
DEV-DIBSTER, line 1: This domain will not be released as this is generic.

domains/settings.json
DEV-DIBSTER, line 1: This domain will not be released as this is generic.

DEV-DIBSTER (changes requested):

PR #10042, opened 2023-12-12
BETA: Register emad_abdou.is-a.dev

EmadAbdou:
Added `emad_abdou.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
We don't support underscores `_` in file names.

PR #10069, opened 2023-12-14
BETA: Register maxgreet.is-a.dev

emmu123:
Added `maxgreet.is-a.dev` using the [dashboard](https://manage.is-a.dev).


emmu123:
[emmu123.github.io/Max-Greet/](https://emmu123.github.io/Max-Greet/)

PR #10102, opened 2023-12-16
Delete scripts/migrate.js

wdhdev:
This script was used for a migration multiple years ago, and has not been used since, I think we should delete it as it won't be used again most likely.


WilliamDavidHarrison:
@phenax

phenax (approved):

PR #10239, opened 2023-12-21
BETA: Register potatooff.is-a.dev

Potatooff:
Added `potatooff.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
We do not support TXT verification or Vercel.

PR #10365, opened 2023-12-24
feat(root): mail records

wdhdev:
I believe we should add mail records to the root domain so maintainers that require an email for is-a.dev purposes can get one. This will only be used for maintainers and not regular users. Only maintainers that require an email will get one.

As of now the maintainers that need an email are just me and @andrewstech. We each require an email for the is-a.dev mail service primarily and the emails will be used so people can forward error reports and other things to our emails for troubleshooting. It will also be used for sending out emails to users if their domain is removed.

We will most likely also setup a support@is-a.dev inbox which will forward to Discord so we can have a more professional email for support.


WilliamDavidHarrison:
@phenax Can you review?

wdhdev:
@phenax

wdhdev:
@phenax Can this be reviewed?

phenax (approved):
LGTM!

PR #10431, opened 2023-12-26
Register Willy.is-a.dev

Willy1011:
Register Willy.is-a.dev with CNAME record pointing to https://github.com/Willy1011/Willy1011.github.io


hackermondev (changes requested):
Invalid `CNAME` on both domains

PR #10444, opened 2023-12-26
BETA: Register xackelon.is-a.dev

xackelon:
Added `xackelon.is-a.dev` using the [dashboard](https://manage.is-a.dev).


hackermondev (changes requested):
You need a `A`/`CNAME` or `MX` record

PR #10453, opened 2023-12-27
Register Beathoven.is-a.dev

PigGoesBrr:
Register Beathoven.is-a.dev with A record pointing to 79.195.11.197:25565.


hackermondev (changes requested):
`A` record cannot have a port

PR #10474, opened 2023-12-28
BETA: Register bit.is-a.dev

apiakuari:
Added `bit.is-a.dev` using the [dashboard](https://manage.is-a.dev).


hackermondev (changes requested):
You need a `A`/`CNAME` record or `MX` record

PR #10507, opened 2023-12-29
BETA: Register root.is-a.dev

legit1337x:
Added `root.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/root.json
+++ b/domains/root.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "legit1337x",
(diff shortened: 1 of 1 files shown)


domains/root.json
DEV-DIBSTER, line 1: This domain name will not be released.

DEV-DIBSTER, line 1: This domain name will not be released.

DEV-DIBSTER (changes requested):

PR #10516, opened 2023-12-30
Register square.is-a.dev

squar3s:
Register square.is-a.dev with CNAME record pointing to https://squar3s.github.io/me/.


hackermondev (changes requested):
`CNAME` records cannot contain slashes, did u mean to use `URL` instead?

PR #10581, opened 2024-01-01
Register n44.is-a.dev

nandhu-44:
Added `n44.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


domains/n44.json
```diff
@@ -0,0 +1,10 @@
+{
+    "owner": {
+        "username": "nandhu-44",
+        "email": "nandhu90040@gmail.com"
+    },
+
+    "record": {
+        "CNAME": "3f83a76d5c9dc21854c1723f._acme.deno.dev."
```
DEV-DIBSTER, line 8: Unable to access site.

The dot is not required at the end.

DEV-DIBSTER (changes requested):

domains/n44.json
nandhu-44: Its not a static site, i just wanted the domain to be having this name. So i was given this cname to be added.

wdhdev: @nandhu-44 It seems to be a ACME challenge? I don't think we can allow this unfortunately.

PR #10653, opened 2024-01-03
Register kelz.is-a.dev

hexaveiii:
Register kelz.is-a.dev with MX record pointing to a.

### DIFF
--- a/domains/kelz.json
+++ b/domains/kelz.json
@@ -1,0 +1,12 @@
+{
+        "owner": {
+           "username": "devkelz",
+           "email": "jakobcameronmuise@gmail.com",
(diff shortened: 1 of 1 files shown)


domains/kelz.json
```diff
@@ -0,0 +1,12 @@
+{
+        "owner": {
+           "username": "devkelz",
+           "email": "jakobcameronmuise@gmail.com",
+           "discord": "803337754925727794"
+        },
+    
+        "record": {
+            "MX": ["a"]
```
DEV-DIBSTER, line 9: Invalid MX record, should be string of values.

DEV-DIBSTER (changes requested):

PR #10684, opened 2024-01-03
Register mcsoft.is-a.dev

prashantdangi:
Register mcsoft.is-a.dev with A record pointing to 3.26.210.137.

### DIFF
--- a/domains/dangiprashant.json
+++ b/domains/dangiprashant.json
@@ -1,0 +1,12 @@
+{
+        "owner": {
+           "username": "dangiprashant",
+           "email": "prashantdangi983@gmail.com",
+           "discord": "994524133847212062"
+        },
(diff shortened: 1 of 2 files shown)


domains/dangiprashant.json
```diff
@@ -0,0 +1,12 @@
+{
+        "owner": {
+           "username": "dangiprashant",
+           "email": "prashantdangi983@gmail.com",
+           "discord": "994524133847212062"
+        },
+    
+        "record": {
+            "A": ["blog"]
```
DEV-DIBSTER, line 9: Invalid A record, should be an array containing strings which contain valid IPv4 addresses.

DEV-DIBSTER (changes requested):

PR #10784, opened 2024-01-06
Changed DNS to GitHub from Cloudflare 

Priyansxu:
## The project doesn't contain index.html file because it is made using nextjs so the page "priyansxu.github.io" is not working you can visit it from https://priyansxu.github.io/profile/


Priyansxu:
Sorry the GitHub page isn't working for nextjs project 🥺

PR #10860, opened 2024-01-07
BETA: Register kenny.is-a.dev

KennySB-dev:
Added `kenny.is-a.dev` using the [dashboard](https://manage.is-a.dev).


FireBoiDiscord:
Do i need to do something else? as i did what the manual instructions said, but the docs link doesn't work.

domains/kenny
```diff
@@ -0,0 +1 @@
+        "email": "kennybugeja@gmail.com"
```
PhoenixOrigin, line 1: This file is invalid; it doesn't have a json file ending and is also invvalid json so please either fix it / delete the file.

PhoenixOrigin (changes requested):

domains/kenny
FireBoiDiscord: i deleted it

PhoenixOrigin: Doesn't show as deleted on here, can you check again please?

wdhdev (approved):

PR #10869, opened 2024-01-08
Update label.yml

wdhdev:
This makes the workflow run when a PR is opened so labels are applied immediately instead of every 10 minutes or so.

### DIFF
--- a/.github/workflows/label.yml
+++ b/.github/workflows/label.yml
@@ -1,13 +1,23 @@
-name: Pull Request Labeler
+name: PR
+
 on:
-  schedule:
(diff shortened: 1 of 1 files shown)


phenax (changes requested):
This was how we did this before but due to a security flaw in pull_request_target that leaks github secrets to forks (which I'm not sure github ever fixed), we decided to remove it. More here: https://github.com/is-a-dev/register/issues/672

wdhdev:
@phenax I didn't personally know about this security exploit, my bad. I might have to start looking into that myself :rofl:, sorry for wasting your time.

PR #10989, opened 2024-01-10
BETA: Register kitsune.is-a.dev

KitsuneDev:
Added `kitsune.is-a.dev` using the [dashboard](https://manage.is-a.dev).


KitsuneDev:
Interesting, so the website creates a PR. Nice work!
Would it be okay to amend the PR with a second subdomain, or are only single domain requests allowed?

DEV-DIBSTER (approved):

PR #11066, opened 2024-01-13
BETA: Register bar-vaibhav.json.is-a.dev

VaibhavSys:
Added `bar-vaibhav.json.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
You don't own json.is-a.dev

PR #11091, opened 2024-01-14
Register surfer.is-a.dev

Divine7576:
Register surfer.is-a.dev with A record pointing to Surfer.

### DIFF
--- a/domains/surfer.json
+++ b/domains/surfer.json
@@ -1,0 +1,12 @@
+{
+        "owner": {
+           "username": "Divine7576",
+           "email": "divine7576@gmail.com",
+           "discord": "1048594307046711417"
+        },
+    
+        "record": {
+            "A": ["surfer"]
+        }
+    }
+    


domains/surfer.json
```diff
@@ -0,0 +1,12 @@
+{
+        "owner": {
+           "username": "Divine7576",
+           "email": "divine7576@gmail.com",
+           "discord": "1048594307046711417"
+        },
+    
+        "record": {
+            "A": ["surfer"]
```
CuteDog5695, line 9: invalid A record

iostpa (changes requested):

domains/surfer.json
Divine7576: What to fill in that?

CuteDog5695: Read here, scroll a bit down and you'll see what A record is for https://www.is-a.dev/docs/domain-structure/

PR #11093, opened 2024-01-14
BETA: Register werewolf .is-a.dev

Fusedon:
Added `werewolf .is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
Please use a valid record and remove the space from the file name.

PR #11138, opened 2024-01-14
BETA: Register lostinvanish .is-a.dev

LostInVanish:
Added `lostinvanish .is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
Please use a valid record and remove the space from your file name.

PR #11175, opened 2024-01-16
Register pi.is-a.dev

ghost:
Added `pi.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).

### DIFF
--- a/domains/pi.json
+++ b/domains/pi.json
@@ -1,0 +1,10 @@
+{
+    "owner": {
+        "username": "FFUV",
+        "email": "Ttrmaniacminecraft@gmail.com"
+    },
+
+    "record": {
+        "MX": ["100"]
+    }
+}


domains/pi.json
```diff
@@ -0,0 +1,10 @@
+{
+    "owner": {
+        "username": "FFUV",
+        "email": "Ttrmaniacminecraft@gmail.com"
+    },
+
+    "record": {
+        "MX": ["100"]
```
DEV-DIBSTER, line 8: Invalid MX records.

DEV-DIBSTER (changes requested):

domains/pi.json
FFUV: How do I do correct mx records

PR #11187, opened 2024-01-16
BETA: Register lifesteal.enderpoint.is-a.dev

EnderPoint07:
Added `lifesteal.enderpoint.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/_serveo-authkey.lifesteal.enderpoint.json
+++ b/domains/_serveo-authkey.lifesteal.enderpoint.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "EnderPoint07",
+        "email": "anshhd64@hotmail.com"
+    },
+    "record": {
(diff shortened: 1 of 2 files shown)


EnderPoint07:
im trynna set up https://serveo.net, its basically a ssh tunnelling service. I thought i could use the custom domain feature with .is-a.dev, is htis not possible?

domains/_serveo-authkey.lifesteal.enderpoint.json
wdhdev, line 1: We do not allow these types of domains.

wdhdev (changes requested):

domains/_serveo-authkey.lifesteal.enderpoint.json
EnderPoint07: okay, but can i ask why we can do the same thing with _discord or _github domains for linking? As all this is doing is linking the site with serveo?

PR #11281, opened 2024-01-19
Update juqwtf.json

juqwtf:
Just add two txt records


wdhdev (changes requested):

PR #11294, opened 2024-01-19
BETA: Register luigbren.is-a.dev

luigbren:
Added `luigbren.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/luigbren.json
+++ b/domains/luigbren.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "luigbren",
+        "email": "luigibrenelli@Gmail.com"
+    },
+    "record": {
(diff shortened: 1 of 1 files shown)


domains/luigbren.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "luigbren",
+        "email": "luigibrenelli@Gmail.com"
+    },
+    "record": {
+        "CNAME": "ns1.a2hosting.com"
```
wdhdev, line 7: We don't support NS records. (Assuming that's what you're trying to add.)

wdhdev (changes requested):

PR #11330, opened 2024-01-20
feat(domain): add email to lylythechosenone.is-a.dev

lylythechosenone:
## Requirements


wdhdev (changes requested):
We don't support CNAME records with other records.

PR #11362, opened 2024-01-22
Register dashboard.almaz.is-a.dev

AlmaziikDev:
Register dashboard.almaz.is-a.dev with A record pointing to 109.71.252.37.


satr14washere (approved):
looks good

PR #11378, opened 2024-01-22
BETA: Register bagelerthefirst.is-a.dev

bageler2:
Added `bagelerthefirst.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
This seems to be a Cloudflare IP address, it cannot be used.

PR #11387, opened 2024-01-23
BETA: Register bo4.is-a.dev

hussain22g:
Added `bo4.is-a.dev` using the [dashboard](https://manage.is-a.dev).


hussain22g:
.

domains/ho2.json
```diff
@@ -0,0 +1,12 @@
+{
+    "owner": {
+        "username": "ho2",
+        "email": "hussainaltahoo944@gmail.com",
+        "repo": "https://ho2.vercel.app/",
+        "description": "ho2 project"
+    },
+    "record": {
+        
+        "TXT": "vc-domain-verify=ho2.is-a.dev,e925e90a6f992a727337"
```
SX-9, line 10: we do not support vercel (assuming since vc-domain-verify is a vercel thing when googled)

SX-9 (changes requested):

hussain22g (commented):
Ok, thanks, now I changed it to cf page

wdhdev (changes requested):
We cannot allow these domains to be registered due to abuse risks.

PR #11429, opened 2024-01-24
Register belsicc.is-a.dev

Belsicc:
Register belsicc.is-a.dev with A record pointing to Belsicc.

### DIFF
--- a/domains/belsicc.json
+++ b/domains/belsicc.json
@@ -1,0 +1,12 @@
+{
+        "owner": {
+           "username": "Belsicc",
+           "email": "yianis.nn@gmail.com",
+           "discord": "1040578981436600330"
+        },
+    
+        "record": {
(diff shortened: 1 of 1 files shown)


domains/belsicc.json
```diff
@@ -0,0 +1,12 @@
+{
+        "owner": {
+           "username": "Belsicc",
+           "email": "yianis.nn@gmail.com",
+           "discord": "1040578981436600330"
+        },
+    
+        "record": {
+            "A": ["belsicc"]
```
DEV-DIBSTER, line 9: Invalid A record, should be IPv4's.

DEV-DIBSTER (changes requested):

domains/belsicc.json
Belsicc: ok

PR #11445, opened 2024-01-24
BETA: Register laman.is-a.dev

Derogater:
Added `laman.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/laman.json
+++ b/domains/laman.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "Derogater",
+        "email": "thelazybloggers@gmail.com"
+    },
+    "record": {
+        "TXT": "txt"
+    }
+}


Derogater:
domain approval request.

domains/laman.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "Derogater",
+        "email": "thelazybloggers@gmail.com"
+    },
+    "record": {
+        "TXT": "txt"
```
SX-9, line 7: txt alone does nothing

satr14washere (changes requested):

domains/laman.json
Derogater: hey, how should i set this up then, can you give an example or you can add as according to the standards on your own accord just keep the name same(in case you are to do it), else please tell me what to do, please.

wdhdev:
https://www.is-a.dev/docs

PR #11453, opened 2024-01-25
BETA: Register vedantnaidu.is-a.dev

whynaidu:
Added `vedantnaidu.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):

PR #11479, opened 2024-01-26
BETA: Register cygg.is-a.dev

ghost:
Added `cygg.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):

PR #11508, opened 2024-01-27
BETA: Register molki.is-a.dev

ghost:
Added `molki.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
We cannot allow these domains to be registered due to abuse risks.

PR #11757, opened 2024-02-02
BETA: Register iarasuster.is-a.dev

iarasuster:
Added `iarasuster.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/iarasuster.json
+++ b/domains/iarasuster.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "iarasuster",
+        "email": "iarasuster@gmail.com"
+    },
+    "record": {
+        "URL": "iarasuster.is-a.dev"
+    }
+}


domains/iarasuster.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "iarasuster",
+        "email": "iarasuster@gmail.com"
+    },
+    "record": {
+        "URL": "iarasuster.is-a.dev"
```
Stef-00012, line 7: Invalid URL record (missing protocol)
also that would cause a redirect loop

Stef-DP (changes requested):

domains/iarasuster.json
iarasuster: what should I change it to?

Stef-00012: the URL of the website you want the domain to redirect to

PR #11763, opened 2024-02-03
BETA: Register efbotlink.is-a.dev

ReincarnatedDeveloper:
Added `efbotlink.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/efbotlink.json
+++ b/domains/efbotlink.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "Heroku403",
+        "email": "juhisharma0757@gmail.com"
+    },
+    "record": {
+        "CNAME": "amorphous-tiger-uqyq9wgl8mh7keh8qonqzxst.herokudns.com"
+    }
+}


domains/efbotlink.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "Heroku403",
+        "email": "juhisharma0757@gmail.com"
+    },
+    "record": {
+        "CNAME": "amorphous-tiger-uqyq9wgl8mh7keh8qonqzxst.herokudns.com"
```
Stef-00012, line 7: Your site is empty, heroku says "There's nothing here, yet."

Stef-DP (changes requested):

domains/efbotlink.json
Heroku403: Yeah becuz I turn off the dyno for future is it a problem so I can switch it back ??

PR #11820, opened 2024-02-04
BETA: Register test1.is-a.dev

PilotTim136:
Added `test1.is-a.dev` using the [dashboard](https://manage.is-a.dev).


iostpa (changes requested):
We are not releasing this domain since it could be abused

PR #11841, opened 2024-02-05
Register cdn.akshtt.is-a.dev

akshtt-dev:
Register cdn.akshtt.is-a.dev with A record pointing to 69.30.249.53.


iostpa (approved):
seems good to me

PR #11878, opened 2024-02-06
BETA: Register daudix.is-a.dev

daudix:
Added `daudix.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/daudix.json
+++ b/domains/daudix.json
@@ -4,6 +4,6 @@
         "email": "ddaudix@gmail.com"
     },
     "record": {
-        "CNAME": "pages.daudix.codeberg.page"
+        "CNAME": "daudix.codeberg.page"
     }
 }
--- a/domains/daudix.json
+++ b/domains/daudix.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "daudix",
+        "email": "ddaudix@gmail.com"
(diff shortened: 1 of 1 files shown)


domains/daudix.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "daudix",
+        "email": "ddaudix@gmail.com"
+    },
+    "record": {
+        "CNAME": "pages.daudix.codeberg.page"
```
Stef-00012, line 7: I cannot reach your site
"Your connection is not private" (and the site uses HSTS so i can't access it using http)

Stef-DP (changes requested):

daudix:
Turns out the .domains were missing, should be working now

domains/daudix.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "daudix",
+        "email": "ddaudix@gmail.com"
+    },
+    "record": {
+        "CNAME": "daudix.codeberg.page"
```
Stef-00012, line 7: I cannot verify the site `daudix.codeberg.page` because it redirects to the is-a.dev domain (it's probably because you already set the CNAME in your server)

PR #11897, opened 2024-02-06
BETA: Register .minecraft.is-a.dev

hey122112:
Added `.minecraft.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/.minecraft.json
+++ b/domains/.minecraft.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "hey122112",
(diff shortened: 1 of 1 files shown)


domains/.minecraft.json
Stef-00012, line 1: file name can not start with a dot (`.`)

Stef-DP (changes requested):

domains/.minecraft.json
wdhdev: Also, we cannot allow it due to impersonation reasons.

PR #11995, opened 2024-02-10
Sorry, it was mistake

ink-developer:
Register Incasator.is-a.dev with A record pointing to test.

### DIFF
--- a/domains/incasator.json
+++ b/domains/incasator.json
@@ -1,0 +1,12 @@
+{
+        "owner": {
+           "username": "Incasat0r",
+           "email": "dimagd2010@gmail.com",
+           "discord": "932881116006875266"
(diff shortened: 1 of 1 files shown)


domains/incasator.json
```diff
@@ -0,0 +1,12 @@
+{
+        "owner": {
+           "username": "Incasat0r",
+           "email": "dimagd2010@gmail.com",
+           "discord": "932881116006875266"
+        },
+    
+        "record": {
+            "A": ["test"]
```
DEV-DIBSTER, line 9: Invalid records, should be strings of valid IPv4s.

DEV-DIBSTER (changes requested):

PR #12003, opened 2024-02-11
BETA: Register ngocnt.is-a.dev

blackmouse572:
Added `ngocnt.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
We do not support Vercel nor TXT verification.

PR #12083, opened 2024-02-14
BETA: Register tcm.is-a.dev

TheCatsMo:
Added `tcm.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/tcm.json
+++ b/domains/tcm.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "TheCatsMo",
+        "email": "real.p.e.f.a@gmail.com"
+    },
+    "record": {
+        "CNAME": "themooingcat.netlify.app"
+    }
+}


domains/tcm.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "TheCatsMo",
+        "email": "real.p.e.f.a@gmail.com"
+    },
+    "record": {
+        "CNAME": "themooingcat.netlify.app"
```
Stef-00012, line 7: Netlify is not suppored

Stef-DP (changes requested):

domains/tcm.json
TheCatsMo: Why?

TheCatsMo: Also, do you have any recommendations as to what host I can use that is supported?

Stef-00012: > Why?

It requires a subdomain of the is-a.dev domain, and if you verify yours, no one else would be able to verify theirs
that's why no one can use netlify (same goes for vercel)

Stef-00012: > Also, do you have any recommendations as to what host I can use that is supported?

if your site is static you can try github pages or cloudflare pages
https://docs.is-a.dev has the guides for some services

PR #12096, opened 2024-02-14
Register aurus.is-a.dev

Ad-AshDev:
Register aurus.is-a.dev with A record pointing to 52.90.95.125:20013.

### DIFF
--- a/domains/aurus.json
+++ b/domains/aurus.json
@@ -1,0 +1,12 @@
+{
+        "owner": {
+           "username": "Ad-AshDev",
+           "email": "adashdev07@gmail.com",
+           "discord": "1169542629415993378"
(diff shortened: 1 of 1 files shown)


domains/aurus.json
```diff
@@ -0,0 +1,12 @@
+{
+        "owner": {
+           "username": "Ad-AshDev",
+           "email": "adashdev07@gmail.com",
+           "discord": "1169542629415993378"
+        },
+    
+        "record": {
+            "A": ["52.90.95.125:20013"]
```
Stef-00012, line 9: Invalid A record (should be only the IP)

Stef-DP (changes requested):

PR #12100, opened 2024-02-14
BETA: Register nevo.is-a.dev

NevoM4:
Added `nevo.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/nevo.json
+++ b/domains/nevo.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "NevoM4",
+        "email": "nevomordechai@gmail.com"
+    },
+    "record": {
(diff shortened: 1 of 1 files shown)


domains/nevo.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "NevoM4",
+        "email": "nevomordechai@gmail.com"
+    },
+    "record": {
```
Stef-00012, line 6: Invalid URL record (missinf protocol)

also you can't just point your domain to google

Stef-DP (changes requested):

PR #12114, opened 2024-02-15
BETA: Register stoltz.is-a.dev

polarxpression:
Added `stoltz.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/stoltz.json
+++ b/domains/stoltz.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "stoltz-dev",
+        "email": "rafaelbraun5@hotmail.com"
+    },
+    "record": {
+        "URL": "stoltz.is-a.dev"
+    }
+}


domains/stoltz.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "stoltz-dev",
+        "email": "rafaelbraun5@hotmail.com"
+    },
+    "record": {
+        "URL": "stoltz.is-a.dev"
```
CuteDog5695, line 7: doesnt have protocol + would create a redirect loop

iostpa (changes requested):

domains/stoltz.json
Stef-00012, line 7: Invalid URL record (missing protocol)
also you would cause a redirect loop with those records

Stef-DP (changes requested):

PR #12186, opened 2024-02-18
BETA: Register kingluc.is-a.dev

kingluc12:
Added `kingluc.is-a.dev` using the [dashboard](https://manage.is-a.dev).


kingluc12 (commented):
URL needs to be changed to https://kingluc.my.id

PR #12214, opened 2024-02-18
BETA: Register kvp.is-a.dev

Samy-E:
Added `kvp.is-a.dev` using the [dashboard](https://manage.is-a.dev).


Omxpro (changes requested):
Site isn't accessible

PR #12265, opened 2024-02-21
Register sans.the.skeleton.is-a.dev

tensura-is-good:
Register sans.the.skeleton.is-a.dev with A record pointing to 172.67.178.60.


Omxpro (changes requested):
cannot access ip

tensura-is-good:
how is iT not  accesible not others please explain what i can do to fix this

PR #12271, opened 2024-02-22
BETA: Register mario-ruoff.is-a.dev

marioruoff:
Added `mario-ruoff.is-a.dev` using the [dashboard](https://manage.is-a.dev).


Omxpro (changes requested):
MX records invalid. you cant use SMTP server of google.

PR #12319, opened 2024-02-23
BETA: Register omairr.is-a.dev

omaiirr:
Added `omairr.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/omairr.json
+++ b/domains/omairr.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "omaiirr",
+        "email": "crazyomairennab@gmail.com"
+    },
+    "record": {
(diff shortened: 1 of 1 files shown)


domains/omairr.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "omaiirr",
+        "email": "crazyomairennab@gmail.com"
+    },
+    "record": {
+        "URL": "omairr.is-a.dev"
```
Stef-00012, line 7: Invalid URL record, missing protocol
also this would cause a redirect loop

Stef-DP (changes requested):

PR #12328, opened 2024-02-23
BETA: Register benchonaut.is-a.dev

benchonaut:
Added `benchonaut.is-a.dev` using the [dashboard](https://manage.is-a.dev).


Omxpro (changes requested):
we cannot review you site. please delete your domain from your github repo so we can access your website

PR #12330, opened 2024-02-24
Register aysko.is-a.dev

aysko-discord:
Register aysko.is-a.dev with CNAME record pointing to relaxed-sable-26f2ce.netlify.app..


Omxpro (changes requested):
we do not support netlify

PR #12352, opened 2024-02-24
Register labber.is-a.dev

ghost:
Added `labber.is-a.dev` using the [CLI](https://www.npmjs.com/package/@is-a-dev/cli).


Omxpro (changes requested):
website took too long to load

Omxpro (changes requested):
cannot use google as a URL record

PR #12367, opened 2024-02-24
Register furry.is-a.dev

furyashnyy:
Register furry.is-a.dev with A record pointing to 31.31.196.104.


Omxpro (changes requested):
incomplete website

VaibhavSys (approved):

PR #12420, opened 2024-02-26
BETA: Register spandan.is-a.dev

spandan11:
Added `spandan.is-a.dev` using the [dashboard](https://manage.is-a.dev).


Omxpro (changes requested):
we do not support vercel

PR #12426, opened 2024-02-26
BETA: Register s409.is-a.dev

S409ISDAD:
Added `s409.is-a.dev` using the [dashboard](https://manage.is-a.dev).


Omxpro (changes requested):
we do not support netlify

PR #12433, opened 2024-02-27
BETA: Register john-paul.is-a.dev

RJohnPaul:
Added `john-paul.is-a.dev` using the [dashboard](https://manage.is-a.dev).


Omxpro (changes requested):
we do not support vercel

PR #12463, opened 2024-02-28
BETA: Register cps_.is-a.dev

CPS7:
Added `cps_.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/cps_.json
+++ b/domains/cps_.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "CPS7",
+        "email": "cpsgaming17@gmail.com"
+    },
+    "record": {
+        "CNAME": "cps_is-a.dev"
+    }
+}


domains/cps_.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "CPS7",
+        "email": "cpsgaming17@gmail.com"
+    },
+    "record": {
+        "CNAME": "cps_is-a.dev"
```
VaibhavSys, line 7: A subdomain cannot point to itself.

VaibhavSys (changes requested):

domains/cps_.json
VaibhavSys: You were trying to make it point to itself, but there is a typo in it.

PR #12560, opened 2024-03-04
BETA: Register test123.is-a.dev

Knightf3chess:
Added `test123.is-a.dev` using the [dashboard](https://manage.is-a.dev).


domains/test123.json
DEV-DIBSTER, line 1: Domain will not be released.

DEV-DIBSTER (changes requested):

PR #12618, opened 2024-03-08
Register api.om.is-a.dev

omsenjalia:
Register api.om.is-a.dev with A record pointing to 69.30.249.53.


iostpa (approved):
Seems good to me

PR #12742, opened 2024-03-14
Fix yunexiz.json

Yunexiz:
I'm having a problem where i get a 502 error no matter what, this only happened after adding a new file to my server, I plan to remove the records, then re-add the records...


Yunexiz:
Messed up...

PR #12841, opened 2024-03-20
BETA: Register tommyc.is-a.dev

FlipMods:
Added `tommyc.is-a.dev` using the [dashboard](https://manage.is-a.dev).


domains/tommyc.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "FlipMods",
+        "email": "tommy.camp@ccsstudent.org"
+    },
+    "record": {
+        "CNAME": ["https://pf-rouge.vercel.app/"]
```
21Z, line 7: 1. Vercel is not supported
2. `CNAME` can't be in an array
3. `CNAME` should not contain protocol (`https://`)

21Z (changes requested):

FlipMods:
What would you recommend that I use instead of vercel?

PR #12889, opened 2024-03-21
Register kowal.is-a.dev

justkowal:
Register kowal.is-a.dev with CNAME record pointing to justkowal.github.io.


iostpa (changes requested):
file is valid and record is valid, but theres like only a "Hello, World!" text. can you add a bit more and then i can merge it?

PR #12893, opened 2024-03-22
Register prajwal.is-a.dev

Toricane:
Register prajwal.is-a.dev with CNAME record pointing to Toricane.github.io.


iostpa (approved):
seems fine to me

PR #12900, opened 2024-03-22
BETA: Register abc.is-a.dev

neohe:
Added `abc.is-a.dev` using the [dashboard](https://manage.is-a.dev).


Stef-DP (approved):
seems good to me

PR #12931, opened 2024-03-23
BETA: Register abdo.is-a.dev

xAbdoAT:
Added `abdo.is-a.dev` using the [dashboard](https://manage.is-a.dev).


Stef-DP (approved):
seems good to me (even if i don't get why u'd need to URL-escape a personal site)

PR #12938, opened 2024-03-23
BETA: Register servers.is-a.dev

Rage65:
Added `servers.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/servers.json
+++ b/domains/servers.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "Rage65",
+        "email": "123thetechguy@gmail.com"
+    },
+    "record": {
+        "CNAME": "rage65.duckdns.org"
+    }
+}


domains/servers.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "Rage65",
+        "email": "123thetechguy@gmail.com"
+    },
+    "record": {
+        "CNAME": "rage65.duckdns.org"
```
CuteDog5695, line 7: site seems to not be accessible

domains/servers.json
CuteDog5695, line 1: will not be released since it has a high risk of being abused

iostpa (changes requested):

domains/servers.json
Rage65: Sorry that is my tailscale adress for my server

domains/servers.json
Rage65: Ok 😔

PR #12952, opened 2024-03-24
BETA: Register jx.is-a.dev

JimsDevelop:
Added `jx.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/jx.json
+++ b/domains/jx.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "JimsDevelop",
+        "email": "jimsdevelop@gmail.com"
+    },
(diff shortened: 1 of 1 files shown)


domains/jx.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "JimsDevelop",
+        "email": "jimsdevelop@gmail.com"
+    },
+    "record": {
+        "A": ["192.168.0.1"]
```
Stef-00012, line 7: you can't use a local IP address

Stef-DP (changes requested):

PR #12983, opened 2024-03-25
BETA: Register idkmaxi.is-a.dev

Molret:
Added `idkmaxi.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/idkmaxi.json
+++ b/domains/idkmaxi.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "Molret",
+        "email": "maximiliano.moldenhauer.r@marygraham.cl"
+    },
+    "record": {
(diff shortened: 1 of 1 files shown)


domains/idkmaxi.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "Molret",
+        "email": "maximiliano.moldenhauer.r@marygraham.cl"
+    },
+    "record": {
+        "CNAME": "ns1.infinityfree.com"
```
Stef-00012, line 7: You can't use namerserver records (NS) and it wouldn't work if u set it as CNAME

Stef-DP (changes requested):

PR #13011, opened 2024-03-26
BETA: Register ratmilk.is-a.dev

ghost:
Added `ratmilk.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/ratmilk.json
+++ b/domains/ratmilk.json
@@ -4,6 +4,6 @@
         "email": "keithpumpkin@gmail.com"
     },
     "record": {
-        "URL": "https://youtu.be/dqw4w9wgxcq"
+        "URL": "https://www.youtube.com/watch?v=dQw4w9WgXcQ"
     }
 }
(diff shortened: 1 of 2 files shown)


domains/ratmilk.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "MilkedRat",
+        "email": "keithpumpkin@gmail.com"
+    },
+    "record": {
+        "URL": "https://youtu.be/dqw4w9wgxcq"
```
Stef-00012, line 7: The video you pointed the domain to is not avaible (private or deleted ig)
"This video isn't available anymore"

Stef-DP (changes requested):

domains/ratmilk.json
Stef-00012: (btw if you want the rickroll, the video ID is `dQw4w9WgXcQ`, you should keep the uppercase ones)

MilkedRat: thats my bad, i copied the wrong link to go to the rickroll, I meant to copy it from the address bar but I did an oopsie instead

PR #13053, opened 2024-03-29
Register julian.is-a.dev

LeoweBlatt13:
Register julian.is-a.dev with CNAME record pointing to ns1.infinityfree.com.

### DIFF
--- a/domains/julian.json
+++ b/domains/julian.json
@@ -1,0 +1,12 @@
+{
+        "owner": {
+           "username": "LeoweBlatt13",
+           "email": "julianwolf2013@gmx.de",
+           "discord": "1222250424309121145"
+        },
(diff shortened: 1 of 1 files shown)


domains/julian.json
```diff
@@ -0,0 +1,12 @@
+{
+        "owner": {
+           "username": "LeoweBlatt13",
+           "email": "julianwolf2013@gmx.de",
+           "discord": "1222250424309121145"
+        },
+    
+        "record": {
+            "CNAME": "ns1.infinityfree.com"
```
Stef-00012, line 9: You can not use CNAME as nameserver (NS)

Stef-DP (changes requested):

PR #13079, opened 2024-03-30
BETA: Register cai.is-a.dev

ukcai696969:
Added `cai.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/cai .json
+++ b/domains/cai .json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "ukcai696969",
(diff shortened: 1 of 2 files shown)


omsenjalia (changes requested):
1. there is a space in the name
2. the ip cannot be `0.0.0.0`

omsenjalia (changes requested):
1. CNAME cannot contain https://
2. We dont support vercel

PR #13081, opened 2024-03-30
Update yunus.json records (hope i did the CNAME right)

Yunexiz:
## Requirements

### DIFF
--- a/domains/yunus.json
+++ b/domains/yunus.json
@@ -4,7 +4,7 @@
         "email": "yunexiz@aol.com"
     },
     "record": {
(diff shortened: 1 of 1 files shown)


omsenjalia (changes requested):
1. CNAME cannot be with MX and TXT records.
2. CNAME cannot be an array.

Yunexiz:
ahh ok..so cf pages wont work with everything else...

TerraPlayz:
> ahh ok..so cf pages wont work with everything else...

Use the A ip array. `/dig` exists.

PR #13085, opened 2024-03-30
BETA: Register creattek.is-a.dev

neopromic:
Added `creattek.is-a.dev` using the [dashboard](https://manage.is-a.dev).


omsenjalia (changes requested):
we dont support `only TXT` records for verification

omsenjalia (changes requested):
1. CNAME cannot have /
2. We do not support vercel

PR #13089, opened 2024-03-30
BETA: Register nx.is-a.dev

ghost:
Added `nx.is-a.dev` using the [dashboard](https://manage.is-a.dev).


domains/nx.json
```diff
@@ -0,0 +1,12 @@
+{
+    "owner": {
+        "username": "jadfoq",
+        "email": "jadfoq@getgoogleoff.me"
+    },
+    "record": {
+    "MX": [
```
DEV-DIBSTER, line 7: Invalid MX records. No TXT records provided? Also these appear to be NS records, could you clarify that these are MX addresses?

DEV-DIBSTER (changes requested):

domains/nx.json
jadfoq: https://github.com/is-a-dev/register/blob/9e02fa2a5590e83d25cd91268f4a6b2d4b81797e/domains/nx.json
https://manage.is-a.dev bug

jadfoq:
Can you use [this file](https://git.disroot.org/jadfoq/register/src/commit/3064471e76feacec904d6afa3cb19cd425ce0425/domains/nx.json), github doesn't work correctly?

jadfoq:
I'm on the [ActivityPub](https://en.wikipedia.org/wiki/ActivityPub) network - https://fe.disroot.org/@jadfoq

CuteDog5695:
> I'm on the [ActivityPub](https://en.wikipedia.org/wiki/ActivityPub) network - https://fe.disroot.org/@jadfoq

we dont support nameservers (also thats not the way to use nameservers, CNAME different to the one you want)

jadfoq:
https://git.disroot.org/jadfoq/register/src/branch/main/domains/nx.json fix, please pull it

domains/nx.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "jadfoq",
+        "email": "jadfoq@getgoogleoff.me"
+    },
+    "record": {
+    "A": "185.31.40.24"
```
CuteDog5695, line 7: A record needs to be in an array + site is not found

iostpa (changes requested):

domains/nx.json
jadfoq: it is a alwaysdata ip https://help.alwaysdata.com/en/sites/use-external-addresses/

PR #13105, opened 2024-03-31
BETA: Register cloudpanel.is-a.dev

TheJ00ker:
Added `cloudpanel.is-a.dev` using the [dashboard](https://manage.is-a.dev).


omsenjalia (changes requested):
it looks like a hosting site but with limited access cant confirm. we dont allow hosting sites.

PR #13113, opened 2024-04-01
BETA: Register koopa.is-a.dev

KoopaCode:
Added `koopa.is-a.dev` using the [dashboard](https://manage.is-a.dev).


omsenjalia (changes requested):
URL records must have a protocol (example `https://`) in front o them

PR #13117, opened 2024-04-01
BETA: Register evangelista.lexi.is-a.dev

ThatDragonixVR:
Added `evangelista.lexi.is-a.dev` using the [dashboard](https://manage.is-a.dev).


domains/evangelista.lexi.json
CuteDog5695, line 1: lexi.is-a.dev seems to not exists so you are not allowed to make a nested subdomain with it

iostpa (changes requested):

PR #13121, opened 2024-04-01
BETA: Register hieu.is-a.dev

hieunotes:
Added `hieu.is-a.dev` using the [dashboard](https://manage.is-a.dev).


omsenjalia (changes requested):
1. CNAME records cannot have `https://`
2. YOU CANT USE NS RECORDS!

PR #13134, opened 2024-04-02
BETA: Register luis.is-a.dev

soyelcreativo:
Added `luis.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/casanova.json
+++ b/domains/casanova.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "appecocab",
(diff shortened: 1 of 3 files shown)


omsenjalia (changes requested):
NS records dont work with CNAME. we do not support NS

omsenjalia (changes requested):
Invalid A record. https://is-a.dev/docs

omsenjalia (changes requested):
Invalid A record

PR #13137, opened 2024-04-02
Register simplymail.is-a.dev

simply-leon:


omsenjalia (changes requested):
CNAME does not work with MX and TXT

domains/simplymail.json
DEV-DIBSTER, line 1: Invalid JSON content.

DEV-DIBSTER (changes requested):

PR #13160, opened 2024-04-03
BETA: Register wumpuss.is-a.dev

Wumpus2000:
Added `wumpuss.is-a.dev` using the [dashboard](https://manage.is-a.dev).


Wumpus2000 (commented):
changed

omsenjalia (changes requested):
Invalid CNAME record. we dont support NS records

PR #13162, opened 2024-04-03
BETA: Register siliconecb.is-a.dev

ghost:
Added `siliconecb.is-a.dev` using the [dashboard](https://manage.is-a.dev).


omsenjalia (changes requested):
we do not support vercel

PR #13165, opened 2024-04-03
BETA: Register mrstickypiston.is-a.dev

MrStickyPiston:
Added `mrstickypiston.is-a.dev` using the [dashboard](https://manage.is-a.dev).


omsenjalia (changes requested):
cannot access website as it redirects to is-a.dev

MrStickyPiston:
Should work now

PR #13177, opened 2024-04-03
BETA: Register kelzz.is-a.dev

hexaveiii:
Added `kelzz.is-a.dev` using the [dashboard](https://manage.is-a.dev).


omsenjalia (changes requested):
invalid

omsenjalia (changes requested):
URL cannot be without a protocol

PR #13192, opened 2024-04-04
BETA: Register haru.ki.is-a.dev

haruki-py:
Added `haru.ki.is-a.dev` using the [dashboard](https://manage.is-a.dev).


haruki-py:
It is fr a proxy

Stef-DP (approved):

PR #13201, opened 2024-04-04
Register ramo.is-a.dev

0xRamo:
Register ramo.is-a.dev with TXT record pointing to test.


Stef-00012:
the readme litterally says to not use the refister site for tests

PR #13447, opened 2024-04-17
Register sono.is-a.dev

LeoweBlatt13:
Register sono.is-a.dev with CNAME record pointing to getsono.github.io.


getSono:
A btw that’s my old acc

PR #13526, opened 2024-04-21
BETA: Register alexandre01.is-a.dev

Alexandre01Dev:
Added `alexandre01.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/alexandre01.json
+++ b/domains/alexandre01.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "Alexandre01Dev",
+        "email": "altata01@hotmail.com"
+    },
+    "record": {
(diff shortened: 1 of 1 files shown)


domains/alexandre01.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "Alexandre01Dev",
+        "email": "altata01@hotmail.com"
+    },
+    "record": {
+        "A": ["127.0.0.1"]
```
Stef-00012, line 7: You can't put a local IP as A record

Stef-DP (changes requested):

PR #13528, opened 2024-04-21
BETA: Register kars.is-a.dev

ghost:
Added `kars.is-a.dev` using the [dashboard](https://manage.is-a.dev).


domains/kars.json
```diff
@@ -0,0 +1,8 @@
+{
+    "owner": {
+        "username": "lean-cc"
+    },
```
Stef-00012, line 4: Missing the email field

Stef-DP (changes requested):

domains/kars.json
lean-cc: fixed

PR #13534, opened 2024-04-21
BETA: Register jobians.is-a.dev

Jobians:
Added `jobians.is-a.dev` using the [dashboard](https://manage.is-a.dev).

### DIFF
--- a/domains/jobians.json
+++ b/domains/jobians.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "Jobians",
+        "email": "jobianstechie@gmail.com"
+    },
+    "record": {
+        "CNAME": "cname.vercel-dns.com"
+    }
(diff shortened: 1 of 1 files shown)


domains/jobians.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "Jobians",
+        "email": "jobianstechie@gmail.com"
+    },
+    "record": {
+        "CNAME": "cname.vercel-dns.com"
```
DEV-DIBSTER, line 7: We do not support Vercel.

DEV-DIBSTER (changes requested):

domains/jobians.json
Jobians: So what will I do?

PR #13546, opened 2024-04-22
BETA: Register korrykatti.is-a.dev

KorryKatti:
Added `korrykatti.is-a.dev` using the [dashboard](https://manage.is-a.dev).


KorryKatti:
i will manually submit it

PR #13556, opened 2024-04-23
Use alias + mx + txt

sometgirldotonline:
## Requirements

### DIFF
--- a/domains/novafox.json
+++ b/domains/novafox.json
@@ -4,7 +4,7 @@
         "email": "foxsdenyt@gmail.com"
     },
     "record": {
-        "ALIAS": "foxsdenyt-github-io.onrender.com",
+        "A":"216.24.57.1",
         "MX": ["mx1.forwardemail.net", "mx2.forwardemail.net"],
         "TXT": "forward-email=foxsdenyt@gmail.com"
     }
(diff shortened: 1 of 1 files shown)


novathefox:
why tf is it complaining abt cname/a records when i dont have them

domains/novafox.json
```diff
@@ -4,6 +4,8 @@
         "email": "foxsdenyt@gmail.com"
     },
     "record": {
-        "CNAME": "foxsdenyt-github-io.onrender.com"
+        "ALIAS": "foxsdenyt-github-io.onrender.com",
```
Stef-00012, line 7: This isn't a valid record that you can use here

Stef-DP (changes requested):

domains/novafox.json
novathefox: ill have to swap it to an A record at "216.24.57.1", which cant be direct access and has to come from a domain, see: https://docs.render.com/configure-other-dns)

domains/novafox.json
```diff
@@ -4,6 +4,8 @@
         "email": "foxsdenyt@gmail.com"
     },
     "record": {
-        "CNAME": "foxsdenyt-github-io.onrender.com"
+        "A":"216.24.57.1",
```
CuteDog5695, line 7: IP Address needs to be in an array

iostpa (changes requested):

PR #13629, opened 2024-04-26
Register ESecondaryschool.is-a.dev

Lukechowns229:
Register ESecondaryschool.is-a.dev with A record pointing to 185.97.217.16.

### DIFF
--- a/domains/esecondaryschool.json
+++ b/domains/esecondaryschool.json
@@ -1,0 +1,12 @@
+{
+        "owner": {
+           "username": "Lukechowns229",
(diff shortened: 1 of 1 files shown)


triach-rold (commented):
Website is reachable but incomplete. Try adding some more stuff to it even if it is a placeholder.

_Il sito web è raggiungibile ma incompleto. Prova ad aggiungere altro, anche se si tratta solo di segnaposto._

iostpa (changes requested):
file is valid, but i would like you to add more stuff to the website

PR #13638, opened 2024-04-27
BETA: Register corenodes.is-a.dev

PM-KIRILL:
Added `corenodes.is-a.dev` using the [dashboard](https://manage.is-a.dev).


wdhdev (changes requested):
We do not allow hosting services to use our domains for advertising purposes.

PR #13714, opened 2024-04-30
BETA: Register nikenhapsari.is-a.dev

nikenhpsr:
Added `nikenhapsari.is-a.dev` using the [dashboard](https://manage.is-a.dev).


MaskDuck (changes requested):
we don't support Netlify :(

PR #13749, opened 2024-05-01
Register janco.is-a.dev

JancoNel:
Register janco.is-a.dev with CNAME record pointing to minecraftkillir.github.io.


triach-rold (changes requested):
Reachable but incomplete. 404.

MaskDuck (changes requested):
inaccessible website

PR #13780, opened 2024-05-02
BETA: Register theta.is-a.dev

ThetaHalo:
Added `theta.is-a.dev` using the [dashboard](https://manage.is-a.dev).


orxngc:
We currently don't support vercel. Your site is also inaccessible.

PR #13792, opened 2024-05-03
BETA: Register scfanime.is-a.dev

scfanime:
Added `scfanime.is-a.dev` using the [dashboard](https://manage.is-a.dev).


triach-rold (approved):
Reachable and complete. Good website, but please add actual content because it looks like a template atm.

PR #13821, opened 2024-05-04
Register proxyping.is-a.dev

ghost:
Register proxyping.is-a.dev with A record pointing to 37.114.35.39.


ghost (commented):
ignore the _discord

PR #13845, opened 2024-05-06
Update crischutu07.json

crischutu07:
Adding ImprovMX for Mail Forwarding 

## Requirements

- [ ] You have completed your website. 
- [ ] You're not using Vercel or Netlify.  
- [ ] There is sufficient information at the `owner` field.

### DIFF
--- a/domains/crischutu07.json
+++ b/domains/crischutu07.json
@@ -1,7 +1,7 @@
 {
   "owner": {
     "username": "crischutu07",
-    "email": "sondolavnxbox.omlet@gmail.com"
+    "email": "crischutu07owo@gmail.com"
   },
   "record": {
     "CNAME": "crischutu07.github.io",
--- a/domains/crischutu07.json
+++ b/domains/crischutu07.json
@@ -4,7 +4,12 @@
     "email": "crischutu07owo@gmail.com"
   },
   "record": {
-    "CNAME": "crischutu07.github.io",
+    "A": [
+      "185.199.108.153",
+      "185.199.109.153",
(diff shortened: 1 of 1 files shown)


domains/crischutu07.json
```diff
@@ -1,9 +1,14 @@
...
+    "username": "crischutu07",
+    "email": "crischutu07owo@gmail.com"
+  },
+  "record": {
+    "CNAME": "crischutu07.github.io",
+    "MX": [
+      "mx1.improvmx.com",
+      "mx2.improvmx.com"
+    ],
+    "TXT": "v=spf1 include:spf.improvmx.com ~all"
```
21Z, line 12: You cant have other records with `CNAME` 
You can use `A` records instead of `CNAME` with four ips pointing to your website

21Z (changes requested):

domains/crischutu07.json
21Z: > Isn't `CNAME` enough for pointing to my GitHub Pages? I think you're confused

CNAME is enough for pointing to github pages, but you can't use CNAME with other records (MX, TXT etc), which is what you're doing here.

CuteDog5695: > Isn't `CNAME` enough for pointing to my GitHub Pages? I think you're confused

you are confused. with CNAME you cant use it with any other records (like MX and TXT), but with A record (which github pages does support) will let you use MX and TXT

iostpa (approved):
seems fine to me now

PR #13861, opened 2024-05-06
Update juqwtf.json

juqwtf:
change record from ip to github pages


iostpa (approved):
seems good to me (site is hosted, but not on the CNAME record)

PR #13933, opened 2024-05-11
Register mc.minmit.is-a.dev

MinMitCoder:
Register mc.minmit.is-a.dev with A record pointing to 141.144.228.165.


iostpa (approved):
seems fine to me

PR #13987, opened 2024-05-15
Register Rouf0x.is-a.dev

Rouf0x:
Register Rouf0x.is-a.dev with CNAME record pointing to rouf0x.github.io.


MaskDuck (approved):
actually not a bad website :) you may have missed your gh verification record but anyways

check the docs

PR #14025, opened 2024-05-18
Register fafu.is-a.dev

delelee:
Register fafu.is-a.dev with CNAME record pointing to faaafu.github.io.


iostpa (changes requested):
Would like you to add more stuff to your website instead of just really short lines (You could add for example an about me)

PR #14033, opened 2024-05-18
Register fishyfish.is-a.dev

ArduFish:
Register fishyfish.is-a.dev with CNAME record pointing to fishyfish.go.ro.

### DIFF
--- a/domains/fishyfish.json
+++ b/domains/fishyfish.json
@@ -1,0 +1,12 @@
+{
+        "owner": {
+           "username": "ArduFish123",
+           "email": "vlad.robert.juganaru@gmail.com",
+           "discord": "619521237777121280"
+        },
+    
+        "record": {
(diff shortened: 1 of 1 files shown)


domains/fishyfish.json
```diff
@@ -0,0 +1,12 @@
+{
+        "owner": {
+           "username": "ArduFish123",
+           "email": "vlad.robert.juganaru@gmail.com",
+           "discord": "619521237777121280"
+        },
+    
+        "record": {
+            "TXT": "amos.ns.cloudflare.com"
```
CuteDog5695, line 9: Wrong records for NS records, also we don't support them at the moment. Even if we did it would only be for maintainers and trusted users most likely.

iostpa (changes requested):

PR #14096, opened 2024-05-23
Delete domains/zwei.json due to abuse risk

21Z:
The site allows anyone to upload images

### DIFF
--- a/domains/zwei.json
+++ b/domains/zwei.json
@@ -1,9 +1,0 @@
-{
-    "owner": {
-        "username": "Vibred",
-        "email": "vibred@pm.me"
-    },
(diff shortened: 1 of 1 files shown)


Vibred:
In fact, I'm the only one using this chart bed. It has an automatic system for recognizing porn pictures, and I manage all the pictures and don't let anyone else store sensitive content. Otherwise it would be blocked by the Chinese firewall and I wouldn't be able to use it myself.

orxngc:
@21Z I understand the concern but I don't think it's fair to take this down just like that, give em' a chance. If it had been reported for abuse previously it would be a different story. Closing.

PR #14111, opened 2024-05-24
Register bin.sono.is-a.dev

getSono:
Register bin.sono.is-a.dev with A record pointing to 69.197.135.202.


iostpa (changes requested):

iostpa (approved):
Seems good to me now

PR #14151, opened 2024-05-26
Register cnicholson.is-a.dev

crnicholson:
Register cnicholson.is-a.dev with CNAME record pointing to cnicholson.is-a.dev.


crnicholson:
Just learned that I can't do it with Vercel.

PR #14228, opened 2024-05-31
Register h3ja.is-a.dev

hejagithub:
Register h3ja.is-a.dev with CNAME record pointing to hejagithub.github.io.


iostpa (approved):
Seems good to me I guess

PR #14233, opened 2024-06-01
Register xlr.is-a.dev

llxlr:
Hi! This is a private repo. It's ok? [llxlr/blog](https://github.com/llxlr/blog)


iostpa (approved):
Seems good to me. (Also URL record is known to have some problems, if the domain doesn't redirect then use https://redirect.pizza)

PR #14286, opened 2024-06-04
Register mrpacific.is-a.dev

MrPacific28:
Register mrpacific.is-a.dev with CNAME record pointing to daphne.ns.cloudflare.com.

### DIFF
--- a/domains/mrpacific.json
+++ b/domains/mrpacific.json
@@ -1,0 +1,12 @@
+{
+        "owner": {
+           "username": "MrPacific28",
+           "email": "mrpacific.n@gmail.com",
+           "discord": "1247600936008941639"
+        },
+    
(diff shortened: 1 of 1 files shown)


domains/mrpacific.json
```diff
@@ -0,0 +1,12 @@
+{
+        "owner": {
+           "username": "MrPacific28",
+           "email": "mrpacific.n@gmail.com",
+           "discord": "1247600936008941639"
+        },
+    
+        "record": {
+            "CNAME": "daphne.ns.cloudflare.com"
```
CuteDog5695, line 9: Wrong way to use NS records + We don't support them at the moment

iostpa (changes requested):

PR #14392, opened 2024-06-10
Register maxh10.is-a.dev

MaxH10Kali:
Register maxh10.is-a.dev with TXT record pointing to Hi, I'm Max..


omsenjalia (changes requested):
Invalid TXT record. https://is-a.dev/docs

PR #14482, opened 2024-06-17
Register odd.is-a.dev

mehedihasanfaiyaz:
Register odd.is-a.dev with CNAME record pointing to www.is-a.dev.


hackermondev (changes requested):
CNAME cannot point at `is-a.dev`. You need to point this to a site you own.

PR #14514, opened 2024-06-20
Register uapi.is-a.dev

NeofetchNpc:
Register uapi.is-a.dev with TXT record pointing to _vercel.

### DIFF
--- a/domains/uapi.json
+++ b/domains/uapi.json
@@ -6,7 +6,7 @@
         },
     
         "record": {
-            "TXT": "_vercel"
+            "TXT": "_vercel.is-a.dev"
         }
     }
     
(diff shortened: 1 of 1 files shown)


domains/uapi.json
```diff
@@ -0,0 +1,12 @@
+{
+        "owner": {
+           "username": "NeofetchNpc",
+           "email": "linucxmd@gmail.com",
+           "discord": "825058927720136754"
+        },
+    
+        "record": {
+            "TXT": "_vercel.is-a.dev"
```
CuteDog5695, line 9: We do not support Vercel currently.

CuteDog5695 (changes requested):

domains/uapi.json
NeofetchNpc: Okay, if that is so

PR #14546, opened 2024-06-22
Revert "Create wacky.json"

DEV-DIBSTER:
Reverts is-a-dev/register#14544


iostpa (approved):
Will merge this since user had their github and netlify records flagged by Google, which is an abuse risk

PR #14576, opened 2024-06-24
Register crystalcoding.is-a.dev

crystalcoding9999:
Register crystalcoding.is-a.dev with CNAME record pointing to crystalcoding9999.github.io.

### DIFF
--- a/domains/crystalcoding.json
+++ b/domains/crystalcoding.json
@@ -1,0 +1,12 @@
+{
+        "owner": {
+           "username": "crystalcoding9999",
+           "email": "crystalcraft9999@gmail.com",
+           "discord": "1102272783712522331"
+        },
+    
+        "record": {
+            "CNAME": "crystalcoding9999.github.io"
+        }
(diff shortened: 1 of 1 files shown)


domains/crystalcoding.json
```diff
@@ -0,0 +1,12 @@
+{
+        "owner": {
+           "username": "crystalcoding9999",
+           "email": "crystalcraft9999@gmail.com",
+           "discord": "1102272783712522331"
+        },
+    
+        "record": {
+            "CNAME": "crystalcoding9999.github.io"
```
Stef-00012, line 9: I can access your website but after a while i get redirected to https://crystalcoding.duckdns.org/ which returns an error caused by too many redirects

Stef-DP (dismissed):

domains/crystalcoding.json
crystalcoding9999: pushed the fix

PR #14585, opened 2024-06-25
Register h3l.is-a.dev

gvalue-04:
Register h3l.is-a.dev with A record pointing to 69.197.135.202.

### DIFF
--- a/domains/h3l.json
+++ b/domains/h3l.json
@@ -1,0 +1,12 @@
+{
+        "owner": {
+           "username": "gvalue-04",
+           "email": "vaporvault04@gmail.com",
+           "discord": "1229820339840618506"
+        },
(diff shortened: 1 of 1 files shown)


gvalue-04:
@orxngc I currently can’t have the website access because I have to create the domain then proxy it.

orxngc:
> @orxngc I currently can’t have the website access because I have to create the domain then proxy it.

In that case please provide a preview of the site, e.g. an image of how it looks on your machine.

gvalue-04:
@orxngc There will be a custom written bio page I will write but a subpage called “projector” will exist, being a modified version of https://share.turbowarp.org , for people to upload and share their turbowarp projects at higher filesize.

PR #14596, opened 2024-06-26
Register lego.is-a.dev

My1Legocraft:
Register lego.is-a.dev with CNAME record pointing to legocraftuk2.pages.dev.


domains/lego.json
CuteDog5695, line 1: Since Lego is a trademark I don't think I can merge this. Maybe you can try another subdomain name?

CuteDog5695 (changes requested):

domains/lego.json
My1Legocraft: It is short for my username, not doing if to any abusing.

PR #14649, opened 2024-06-30
Register zedxos.is-a.dev

zedxos:
Register zedxos.is-a.dev with CNAME record pointing to cheerful-monstera-e4af4f.netlify.app.


iostpa (dismissed):

iostpa (approved):
Now it seems fine

PR #14695, opened 2024-07-03
Register lightsys32.is-a.dev

not-amplify:
Register lightsys32.is-a.dev with A record pointing to 104.243.45.193.


not-amplify:
> Inaccessible site. If it's not possible to be up right now, please provide a preview of the site instead.

we have an accessible preview at https://lightgo.app

PR #14698, opened 2024-07-03
Register grbhost.is-a.dev

Gustag123rb1:
Register grbhost.is-a.dev with TXT record pointing to @.


orxngc (changes requested):
You cannot have just TXT records. Read the documentation next time before making a PR.

PR #14709, opened 2024-07-05
Register painscreen.is-a.dev

iHategithub9000:
Register painscreen.is-a.dev with CNAME record pointing to painscreenisadev.glitch.me.


iHategithub9000:
links should work now

PR #14725, opened 2024-07-06
Register alecks.is-a.dev

Alecks20:
Register alecks.is-a.dev with CNAME record pointing to euphoria.alecks.dev.

### DIFF
--- a/domains/alecks.json
+++ b/domains/alecks.json
@@ -1,0 +1,12 @@
+{
+        "owner": {
+           "username": "Alecks20",
+           "email": "109296824+Alecks20@users.noreply.github.com",
+           "discord": "612522818294251522"
+        },
+    
+        "record": {
+            "CNAME": "euphoria.alecks.dev"
+        }
+    }
+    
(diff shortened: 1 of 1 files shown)


domains/alecks.json
```diff
@@ -0,0 +1,12 @@
+{
+        "owner": {
+           "username": "Alecks20",
+           "email": "109296824+Alecks20@users.noreply.github.com",
```
CuteDog5695, line 4: Either remove your no-reply email or replace it with your own.

iostpa (changes requested):
Tried visiting the page but I can't since the dev domain doesn't have SSL so I had to check ReviewMate which seems to give page 404 not found. Can you give a preview of your website?

Alecks20:
The website currently returns a 404 its an nginx reverse proxy, once the domain is initialized I will setup a lets encrypt cert. 

The website I'm gonna deploy onto it is also accessible at https://alecks.dev

domains/alecks.json
Alecks20: Completed this

iostpa (approved):
Seems good to me now

PR #14748, opened 2024-07-07
Update alecks.json (Change CNAME target)

Alecks20:
Moved my website to another server


Alecks20:
If its a different maintainer reviewing than last time, look at https://github.com/is-a-dev/register/pull/14725#issuecomment-2211706686, thats why the website preview says "404 not found"

PR #14768, opened 2024-07-09
Register arvindt.is-a.dev

ARVIN3108:
Register arvindt.is-a.dev with CNAME record pointing to arvin3108.github.io & Discord Domain Verification


ARVIN3108 (commented):
Done, i made it already.

image

PR #14770, opened 2024-07-09
Domain

x83owen:


domains/_github-pages-challenge-owentechv.owentech.json
Stef-00012, line 1: soowentech doesn't exist as is-a.dev domain so you can'y make a subdomain of it

Stef-DP (dismissed):

iostpa (approved):
Seems good to me now

PR #14818, opened 2024-07-11
Register neca.is-a.dev

nekotamogas:
Register neca.is-a.dev with TXT record pointing to zoho-verification=zb18547320.zmverify.zoho.eu.

### DIFF
--- a/domains/neca.json
+++ b/domains/neca.json
@@ -1,0 +1,12 @@
+{
+        "owner": {
+           "username": "nekotamogas",
+           "email": "dazzerdev@gmail.com",
+           "discord": "797458672225091594"
+        },
+    
+        "record": {
+            "TXT": "zoho-verification=zb18547320.zmverify.zoho.eu"
+        }
+    }
+    


domains/neca.json
```diff
@@ -0,0 +1,12 @@
+{
+        "owner": {
+           "username": "nekotamogas",
+           "email": "dazzerdev@gmail.com",
+           "discord": "797458672225091594"
+        },
+    
+        "record": {
+            "TXT": "zoho-verification=zb18547320.zmverify.zoho.eu"
```
CuteDog5695, line 9: I think you need more records than TXT for zoho (like MX records)

CuteDog5695 (changes requested):

domains/neca.json
nekotamogas: It only gave me TXT record and i can choose also between CNAME or something with uploading file on the website.

CuteDog5695 (approved):
I will merge this so that you can move on with zoho's instructions.

PR #14825, opened 2024-07-11
Register FanaticExplorer.is-a.dev

FanaticExplorer:
Register FanaticExplorer.is-a.dev with A record pointing to 192.168.0.1.


FanaticExplorer:
Yea, I set 192.168.0.1, because I just want to test how it works, and I don't have server yet

If that is wrong, and can't be redacted by myself later, better reject it XD

FanaticExplorer:
Got it, thanks, sorry for disturbing

PR #14885, opened 2024-07-14
Gain contributor by changing the Dockerfile a lot.

Yunexiz:
Hello, i have made some critical updates to the Dockerfile in hopes of gaining myself the contributor role in is-a.dev. I have noticed some highly dangerous oversights which is the fact that newbie coders wont be able to understand the code, so I have added amazing comments, to make sure everyone is included in this journey to the end of the domain name system.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,29 @@
+# Uses Ubuntu 20.04 LTS
 FROM ubuntu:20.04
 
 ENV TERM xterm
 RUN ln -fs /usr/share/zoneinfo/America/New_York /etc/localtime
 
+# Run the commands apt-get update, apt-get install
 RUN apt-get -y update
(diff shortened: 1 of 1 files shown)


QuinceTart10:
I personally think that the Containerfile is readable enough as is, and that your comments made basically no difference.

Yunexiz:
> I personally think that the Containerfile is readable enough as is, and that your comments made basically no difference.

Thats kind of the point, i dont want to ruin anything so ive just made useless comments. My only goal was the contributor role.

wdhdev:
The contributor role isn't really given based off commits in this repo and it's only given to users that have contributed significant amounts of code and effort in to helping the service operate. (e.g. the Discord bots). This PR will not get you that role and also it is kind of pointless, we don't need conments for the Dockerfile, it is pretty much self-explanatory as @QuinceTart10 said above.

PR #14919, opened 2024-07-15
remove domain using cyclic 

wdhdev:
@catfoolyou I'm not sure if you are aware but Cyclic (the provider your domain is linked to) shutdown a few months ago as seen in their blog post: https://www.cyclic.sh/posts/cyclic-is-shutting-down/

If you are okay with us just removing your domain as it is leading nowhere just comment/approve these changes, or if you want to change providers or something you can do that too. You can of course re-register your domain at any point :)

I'll remove the domain in ~48 hours if there isn't a response, but you will still be able to re-register it in the future.

### DIFF
--- a/domains/catfoolyou.json
+++ b/domains/catfoolyou.json
@@ -1,9 +1,0 @@
-{
-  "owner": {
-    "username": "catfoolyou",
-    "email": "catfoolyou.github@manyme.com"
-  },
-  "record": {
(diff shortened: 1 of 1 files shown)


DEV-DIBSTER (approved):

catfoolyou:
Sure. Go ahead and remove it, I have no use for it anymore anyways.

PR #14937, opened 2024-07-16
changing the alt text for the discord server link in README.md

xkcdnerd:
pls i really wanna be a contributer


iostpa (changes requested):
Sorry but this is actually useless code. Also I don't think just editing one useless line will get you Contribuiter role in the discord server.

CuteDog5695:
I'll close this PR since it changes almost nothing

wdhdev:
> Sorry but this is actually useless code. Also I don't think just editing one useless line will get you Contribuiter role in the discord server.

It won't. https://github.com/is-a-dev/register/pull/14885#issuecomment-2227250839

PR #14954, opened 2024-07-17
Register xkotelek.is-a.dev

xKotelek:
Register xkotelek.is-a.dev with CNAME record pointing to xkotelek.github.io.

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,1 +1,0 @@
-xkotelek.is-a.dev
--- a/CNAME
+++ b/CNAME
@@ -1,0 +1,1 @@
+xkotelek.is-a.dev
--- a/domains/kotelek.json
+++ b/domains/kotelek.json
@@ -1,9 +1,0 @@
-{
-  "owner": {
-    "username": "xKotelek",
-    "email": "contact@kotelek.dev"
-  },
-  "record": {
-    "CNAME": "kotelek.dev"
-  }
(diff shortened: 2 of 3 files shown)


CNAME
CuteDog5695, line 1: Delete this file please

iostpa (changes requested):

CNAME
xKotelek: what file?

xKotelek: kotelek.json was deleted

iostpa (approved):
Seems good to me now

xKotelek:
Still can't enforce HTTPS 
image
And an error:
image
Connection is not private
Attackers may be trying to steal your information from xkotelek.is-a.dev (such as passwords, messages, or credit card information). More information
net::ERR_CERT_COMMON_NAME_INVALID
xkotelek.is-a.dev typically uses encryption to protect your information. When Brave tried to connect to xkotelek.is-a.dev this time, it sent back unusual and incorrect login information. This can happen when an attacker is impersonating xkotelek.is-a.dev or when the Wi-Fi login screen interrupts the connection. Your information is still safe because the connection in Brave ended before any data was exchanged.

You can't visit xkotelek.is-a.dev right now because it uses HSTS. Network errors and attacks are usually temporary, so the site will likely be available later.

PR #15014, opened 2024-07-20
Register formuna.is-a.dev

FormunaGit:
Register formuna.is-a.dev with TXT record pointing to vc-domain-verify=formuna.is-a.dev,13260821808507d1277b.

### DIFF
--- a/domains/formuna.json
+++ b/domains/formuna.json
@@ -1,0 +1,12 @@
+{
+        "owner": {
+           "username": "FormunaGit",
+           "email": "",
+           "discord": "754657845563097108"
+        },
+    
+        "record": {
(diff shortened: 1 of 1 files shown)


FormunaGit:
Um, I'm a bit new to URL stuff like this so if needed, the thing that vercel gave me is this:
Type    Name    Value
TXT      _vercel   vc-domain-verify=formuna.is-a.dev,13260821808507d1277b

iostpa (changes requested):
We do not support Vercel.

FormunaGit:
What? Why?

FormunaGit:
Oh, nevermind. I'll use another provider then

PR #15160, opened 2024-07-28
Register mail.amaankazi.is-a.dev for emails

Amaan-Kazi:
Registering mail.amaankazi.is-a.dev for emails
amaankazi.is-a.dev is already registered by me

## Requirements

- [x] You're not using Vercel.  


Yunexiz:
Since you're using ImprovMX, I think you should consider scrapping the mail nested domain (subdomain on a subdomain). You could just set it to accept mail from mail@amaankazi.is-a.dev instead. mail@mail.amaazkazi.is-a.dev doesn't really make sense. Just a suggestion, but think about it. You could also let is just accept mail from any tag (Not just mail@<domain>) since it doesn't really affect the website.

omsenjalia (approved):

Amaan-Kazi:
> Since you're using ImprovMX, I think you should consider scrapping the mail nested domain (subdomain on a subdomain). You could just set it to accept mail from [mail@amaankazi.is-a.dev](mailto:mail@amaankazi.is-a.dev) instead. [mail@mail.amaazkazi.is-a.dev](mailto:mail@mail.amaazkazi.is-a.dev) doesn't really make sense. Just a suggestion, but think about it. You could also let is just accept mail from any tag (Not just mail@) since it doesn't really affect the website.

i tried doing that but i am already using amaankazi.is-a.dev for my website using CNAME record
when i tried adding MX and TXT records, a moderator told me i cant add other records with CNAME.
so i dont think theres anyway for me to have both the site and the mail be on the same subdomain since im using coludflare pages for my website and it only has CNAME record

PR #15183, opened 2024-07-29
Register boink.is-a.dev

PogMaster9001:
Register boink.is-a.dev with CNAME record pointing to projectdevs.net.

### DIFF
--- a/domains/amd.json
+++ b/domains/amd.json
@@ -1,0 +1,11 @@
+{
+        "owner": {
+           "username": "PogMaster9001",
+           "email": "",
+           "discord": "1190052608346435597"
+        },
+
+        "record": {
+            "CNAME": "projectdevs.net"
+        }
+    }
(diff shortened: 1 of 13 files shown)


domains/i.json
wdhdev, line 1: This domain will be reserved for future use, sorry.

domains/intel.json
wdhdev, line 1: Due to copyright you cannot register this domain.

domains/nvidia.json
wdhdev, line 1: Due to copyright you cannot register this domain.

domains/x.json
wdhdev, line 1: Due to copyright you cannot register this domain. (X/Twitter)

domains/amd.json
wdhdev, line 1: Due to copyright you cannot register this domain.

wdhdev (changes requested):
There is no way you are actually going to be using all these domain names.

PR #15270, opened 2024-07-31
Register mateusz.is-a.dev

devsuperking:
Register mateusz.is-a.dev with TXT record pointing to vc-domain-verify=mateusz.is-a.dev,61aefe378836b77e2c72.


JustDeveloper1:
i think txt records alone are do not do anything

MaskDuck (changes requested):

PR #15364, opened 2024-08-05
Register kokoapi.is-a.dev

kokofixcomputers:
Register kokoapi.is-a.dev with A record pointing to 45.90.12.32.

### DIFF
--- a/domains/kokoapi.json
+++ b/domains/kokoapi.json
@@ -1,0 +1,13 @@
+{
+        "owner": {
+           "username": "kokofixcomputers",
+           "email": "",
+           "discord": "1096839213313446019",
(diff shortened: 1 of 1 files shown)


kokofixcomputers:
Screenshot 2024-08-05 at 10 11 45 AM
Do not think this is inaccessible just by pasting the ip in the browser!!!!! I have to manually add this subdomain into my hosting service later as right now it says That domain does not point to this server. Make sure your A record is set to 45.90.12.32. and doesn't let me add

kokofixcomputers:
# Read above before declining ^^^^^^^^^^^^^

iostpa (approved):
Seems good to me

kokofixcomputers:
Thanks!

PR #15443, opened 2024-08-08
Register nvrxstdwrth.is-a.dev

nxstd:
Register nvrxstdwrth.is-a.dev with A record pointing to 65.21.61.55.

### DIFF
--- a/domains/nvrxstdwrth.json
+++ b/domains/nvrxstdwrth.json
@@ -1,0 +1,13 @@
+{
+        "owner": {
+           "username": "HIlight3R",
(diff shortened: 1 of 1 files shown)


HIlight3R:
This is the PHPMyAdmin panel. I need access to it at all times, but right now my ISP (or router, which I don't have access to) blocks unsecured HTTP connections. I can only get a security certificate for the domain :(

HIlight3R:
Damn, wrong port mapping. You must have seen the database admin panel. Check again, there should be a dashboard there now.

PR #15444, opened 2024-08-08
Register ena0.is-a.dev

ena0:
Register ena0.is-a.dev with TXT record pointing to test.


oliik2013:
Don't make test pulls, it wastes Maintainers time

PR #15468, opened 2024-08-09
Register rager1.is-a.dev

Antsbatscats:
Register rager1.is-a.dev with CNAME record pointing to https://rager1-web.vercel.app/.


oliik2013:
Vercel and https:// and / is not allowed in CNAME, Vercel is not allowed on this platform completly

PR #15628, opened 2024-08-17
Register os.is-a.dev

heyturkiye204:
Register os.is-a.dev with CNAME record pointing to cdn-rsdv-hey.netlify.app


iostpa (changes requested):
I don't think we allow www.subdomain (could be wrong though)

domains/www.heytr.json
CuteDog5695, line 1: Delete this file.

domains/os.json
CuteDog5695, line 1: Just a comment, but im not sure if the name is allowed (though im thinking like this since im a bit picky)

iostpa (changes requested):

iostpa (approved):
I guess its fine now

PR #15631, opened 2024-08-18
Register de.mcstatus.is-a.dev

EducatedSuddenBucket:
Register de.mcstatus.is-a.dev with A record pointing to 216.24.57.1.


EducatedSuddenBucket:
same things as mcstatus.is-a.dev but hosted in diffrent locations in render

Stef-DP (approved):

PR #15645, opened 2024-08-19
Register feliciaxxx.is-a.dev

FeliciaXxxTop:
Register feliciaxxx.is-a.dev with TXT record pointing to dh=017b0e9fa94bd56685b9ff23133c382ec88fd899.

### DIFF
--- a/domains/felicia.json
+++ b/domains/felicia.json
@@ -1,0 +1,13 @@
+{
+        "owner": {
+           "username": "FeliciaXxxTop",
+           "email": "",
+           "discord": "801672371588956170",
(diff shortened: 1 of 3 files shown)


domains/feliciaxxx.json
orxngc, line 1: This should be under a _discord domain. Read the [documentation](https://www.is-a.dev/docs/discord-verification/).

domains/feliciaa.json
orxngc, line 1: TXT records alone aren't allowed.

domains/felicia.json
orxngc, line 1: This file is fine. The other two still have the same problems.

orxngc: If ya want me to delete the other two if you don't know how, I can.

PR #15654, opened 2024-08-19
Register luchii.is-a.dev

maxwalks:
Register luchii.is-a.dev with A record pointing to 76.76.21.21.


maxwalks:
If you add that A record, its an ip of vercel but vercel makes sure it's my
project that will get displayed. It should just work if you add that ip
address.

orxngc:
Regardless, we do not support Vercel. Closing as stale.

PR #15760, opened 2024-08-26
Register cdn.is-a.dev

KoopaCode:
Register cdn.is-a.dev with A record pointing to 45.126.211.69.


wdhdev:
Too generic for a root domain.

PR #15779, opened 2024-08-26
Register lowsummer.is-a.dev New

aglsk:
Register lowsummer.is-a.dev with TXT record pointing to lowsummer.

### DIFF
--- a/domains/lowsummer.json
+++ b/domains/lowsummer.json
@@ -3,8 +3,8 @@
     "repo": "https://github.com/LowSummer/lowsummer.github.io",
     "owner": {
         "username": "LowSummer",
-        "email": "<email@address>",
-        "twitter": "<twitter-username>"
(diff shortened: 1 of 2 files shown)


domains/lowsummer.json
```diff
@@ -0,0 +1,12 @@
+{
+    "description": "Site Low Summer",
+    "repo": "https://github.com/LowSummer/lowsummer.github.io",
+    "owner": {
+        "username": "LowSummer",
+        "email": "<email@address>",
+        "twitter": "<twitter-username>"
```
DEV-DIBSTER, line 7: Socials are invalid, provide ones that have valid entries. You can use owl, email, or another valid way of contact.

DEV-DIBSTER (changes requested):

domains/lowsummer.json
aglsk: I will correct it

PR #15821, opened 2024-08-29
Register mr-julus.is-a.dev

BougSoon-Studio:
Register mr-julus.is-a.dev with CNAME record pointing to mr-julus.web.app.


domains/mr-julus
Stef-00012, line 1: The file should be a JSON fil (end with `.json`)

Stef-DP (dismissed):

domains/mr-julus
BougSoon-Studio: it's done

PR #15861, opened 2024-09-01
Register maki.is-a.dev

ghost:
Register maki.is-a.dev with CNAME record pointing to mackarroll.pages.dev.


iostpa (dismissed):
Uh can you add more content to the site?

iostpa (approved):
Seems good to me now

PR #15870, opened 2024-09-02
Register nate.is-a.dev

natejw:
Register nate.is-a.dev with A record pointing to 172.66.0.70.


natewombwell:
image
here you go sorry about that. carrd doesn't allow direct access to domains.

PR #15925, opened 2024-09-04
Removing domains with Vercel TXT verification records

iostpa:


VaibhavSys (approved):

VaibhavSys:
Hello @drizion and @CcNicebruh

We will be removing these two unused domains with invalid records in 24 hours to test Vercel Integration, you may respond to this PR to avoid deletion and modify the records to something else. You will still be able to re-register these domains later.

CuteDog5695:
Merging PR since its been 24 hours. Even though the domains are being removed, you can re-register the domain but this time without the Vercel TXT record.

PR #15933, opened 2024-09-05
Register www.haseeb.is-a.dev

Haseebprivate:
Register www.haseeb.is-a.dev with CNAME record pointing to haseeb-khalid.netlify.app.


domains/www.haseeb.json
CuteDog5695, line 1: As I recall, we don't support `www` nested subdomains.

iostpa (dismissed):

PR #15979, opened 2024-09-08
Register _acme-challenge.us.mcstatus.is-a.dev

EducatedSuddenBucket:
Register _acme-challenge.us.mcstatus.is-a.dev with CNAME record pointing to e6797667cd820de372823eb4._acme.deno.dev.

PR #15981, opened 2024-09-08
ci: fix CNAME validations for ACME records

MaskDuck:
this PR is brought about by #15979

image


wdhdev:
yeah its kinda broken mate

wdhdev (changes requested):
Can we just allow underscores entirely, instead of just _acme?

wdhdev (approved):

PR #16000, opened 2024-09-08
Register _discord.raccoon.is-a.dev

raconmeow:
Register _discord.raccoon.is-a.dev with TXT record pointing to dh=e9381b84c8007443b186820d32a3797bd8fa537d.


domains/_discord.nils.json
Stef-00012, line 1: the nils domain doesn't seem to be owned by you

Stef-DP (changes requested):

domains/_discord.nils.json
nilsraccoon: oops i forgot to remove that commit

Stef-DP (approved):

nilsraccoon:
ty

PR #16020, opened 2024-09-10
Register arta.is-a.dev

artaabedi:
Register arta.is-a.dev with CNAME record pointing to artaabedi.github.io.


Stef-DP (dismissed):
You can not have multiple domains pointing to the same domain, on github pages you can only link one domain to a website

VaibhavSys (approved):

PR #16073, opened 2024-09-13
Test to see if Vercel is actually supported now

iostpa:


CuteDog5695:
Forgot that we didn't add it to the list of allowed underscores, whoops

PR #16192, opened 2024-09-19
Register _acme-challenge.api.juststudio.is-a.dev

ghost:


_acme-challenge.api.juststudio.json
Stef-00012, line 1: The file should be in the domains folder

Stef-DP (dismissed):

PR #16194, opened 2024-09-19
Update shy.json redirect

ImShyMike:
Changing back to redirect.pizza because the URL record decided to not work anymore

### DIFF
--- a/domains/shy.json
+++ b/domains/shy.json
@@ -6,7 +6,7 @@
         },
     
         "record": {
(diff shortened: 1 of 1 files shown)


Stef-DP (approved):

MohammedKHC0:
Hello Guys, Iam also using url record, But it works well, So how exactly it's not working any more?

Am i at the safe zone by using URL record or should i use redirect.pizza

PR #16220, opened 2024-09-21
Update satr14.json (URL redirect)

satr14washere:
all good


iostpa (approved):
If it doesnt redirect then use redirect.pizza

PR #16260, opened 2024-09-24
Fix cname record for forum.cupglass.is-a.dev

ghost:
Sorry, i didnt read it carefully. So instead of
`cname.freeflarum.com`
it SHOULD point to that domain WITH a dot at the end, like this
`cname.freeflarum.com.`
https://docs.freeflarum.com/en/tutorials/custom-domain/

(even the checks are throwing an error because of the dot, but i didnt expect this to be required, a single dot???)

## Requirements

### DIFF
--- a/domains/forum.cupglass.json
+++ b/domains/forum.cupglass.json
@@ -9,6 +9,6 @@
         "discord": "cupglassdev"
     },
     "record": {
-        "CNAME": "cname.freeflarum.com"
+        "CNAME": "cname.freeflarum.com."
     }
 }


DEV-DIBSTER:
@cupglassDEV You're records are correct from before.

cupglassDEV:
> @cupglassDEV You're records are correct from before.

But it says i need to point with a dot, because when i tried it this happens

image

DEV-DIBSTER:
> But it says i need to point with a dot, because when i tried it this happens

We already handle the "." at the end, so you're records from before should have already propagated. You can also see that the validation fails when you have the "." per our DNS rules.

cupglassDEV:
> > But it says i need to point with a dot, because when i tried it this happens
> 
> We already handle the "." at the end, so you're records from before should have already propagated. You can also see that the validation fails when you have the "." per our DNS rules.

Oh okay, thanks for the help! I will try to point it again on the forum settings
Thank you :)

EducatedSuddenBucket:
@cupglassDEV that is issue from freeflarum due to dns propagation, domains without and with  work same for eg both google.com and google.com. will work on a browser. I have used deno with is-a.dev they also gave cname with dots but it works even in the json I didn't add dot in the cname

cupglassDEV:
> @cupglassDEV that is issue from freeflarum due to dns propagation, domains without and with work same for eg both google.com and google.com. will work on a browser. I have used deno with is-a.dev they also gave cname with dots but it works even in the json I didn't add dot in the cname

Oh, thanks for the reply! <sup>I might try to remove this json file soon 😉™️</sup>

PR #16326, opened 2024-09-27
Register eric-.is-a.dev

JustAnEric:
Register eric-.is-a.dev with CNAME record pointing to ericisadev.rf.gd.


domains/eric-.json
DEV-DIBSTER, line 1: Domain file name is invalid.

DEV-DIBSTER (changes requested):

PR #16349, opened 2024-09-28
Register raikan-san.is-a.dev

wrothmir:
Register raikan-san.is-a.dev with CNAME record pointing to raikan-san.github.io.


iostpa (approved):
Seems good to me

PR #16394, opened 2024-10-01
Register visualstudios.is-a.dev

urasdbwd:
Register visualstudios.is-a.dev with A record pointing to 193.149.164.209.


domains/visualstudios.json
CuteDog5695, line 1: I don't think we can allow this domain since its almost identical to "Visual Studio" which is a very popular IDE and its most probably trademarked.

iostpa (changes requested):

PR #16453, opened 2024-10-06
Register support.is-a.dev

ngcongtunglam209:
Register ask-google.is-a.dev with A record pointing to 172.217.16.238.


domains/support.json
orxngc, line 1: support is not a domain name we'll allow

ngocdiep2006: okay can i reg with ask-google?

PR #16486, opened 2024-10-08
Register mkeko.is-a.dev

snowwysillycat:
Register mkeko.is-a.dev with TXT record pointing to vc-domain-verify=mkeko.is-a.dev,060738e86128a300ceee.


iostpa (changes requested):
Not the right way to use Vercel, please go here to see what you are supposed to do: https://www.is-a.dev/docs/vercel/

iostpa (approved):
Seems good to me now

PR #16514, opened 2024-10-10
Register rimikocommunity.is-a.dev

79SUBTEAM:
Register rimikocommunity.is-a.dev with CNAME record pointing to noithattoha.com.


iostpa (approved):
Seems fine to me ig, we've merged PRs like this before

PR #16587, opened 2024-10-14
Register EZDevelopment.is-a.dev

LootingVI:
Register EZDevelopment.is-a.dev with A record pointing to 2.56.244.116.


LootingVI:
It is complete?

LootingVI:
For my last domain it worked

LootingVI:
I have no desire for a "real" site because here in Germany we have very strict rules about that.

orxngc:
> For my last domain it worked

Somebody merged a PR with a panel? That's not allowed and shouldn't have been merged, unless it was a nested subdomain.



> I have no desire for a "real" site because here in Germany we have very strict rules about that.

A "Hi, I'm orangc and my Discord user ID is 961063229168164864" will suffice as complete per our standards. German laws surely don't prohibit something like that. Afterwards you could add your panel as a nested subdomain to this domain.

LootingVI:
It's not that it's forbidden, but in Germany we have an imprint obligation for every public website. This means that you are formally forced to give out private data such as telephone number, place of residence etc. so that the police can contact you if there is something illegal with the site. Would it be possible to get 2 nested domains? So panel.ezdevelopment.is-a.dev and MC.ezdevelopment.is-a.dev?

orxngc:
> Would it be possible to get 2 nested domains? So panel.ezdevelopment.is-a.dev and MC.ezdevelopment.is-a.dev?

Yes, as long as you own ezdevelopment.is-a.dev.

PR #16614, opened 2024-10-16
Register blooket.is-a.dev

xgeneric:
Register blooket.is-a.dev with CNAME record pointing to https://xgeneric.github.io/blooket/.


wdhdev:
Name infringes on a trademark.

PR #16646, opened 2024-10-18
Username and site name change

mystichronicle:
Username changed from debjit-mandal to tensorgenius and same for the site

## Website Link/Preview
https://tensorgenius.github.io


iostpa (approved):
Seems good to me

PR #16683, opened 2024-10-19
Register cleak.is-a.dev

Sife-Cleak:
Added `cleak.is-a.dev`


Satindar31:
You got NS records so you're also gonna have to provide a reason.

Sife-Cleak:
I want to use the domain for email address and I think I'll change DNS many times due to many sites I want to add to it.Could I use the NS record?

orxngc:
not a good enough reason so no

PR #16691, opened 2024-10-19
Update luoxh.json

Melishy:
I know i have added lots of CNAME, but i want to try clerk for it. I changed my email address also.


Melishy:
welp understand its only allowed for 1 cname 😭

PR #16708, opened 2024-10-20
Update luoxh.json

Melishy:
## commit message: beautify, add txt records & discord, and change the email


Melishy:
whatever i don't know about records

PR #16713, opened 2024-10-20
Register gameserver.ardaninho.is-a.dev

Ardaninho:
Register gameserver.ardaninho.is-a.dev with A record pointing to 207.244.199.139.


omsenjalia (changes requested):
Cannot access website

omsenjalia (approved):

PR #16760, opened 2024-10-23
Create samp.alora.json

aloramiaa:
This PR is for Hosting a Samp Server 
San Andreas Mulitplayer - https://www.sa-mp.mp/
To make my  SA-MP server accessible via a custom domain name

The Samp Server is Currently Running on 

IP: 46.183.184.33
Port: 3055

I want to make the Ip as samp.alora.is-a.dev:3055

Thank You.


iostpa (approved):
I guess its fine

PR #16832, opened 2024-10-26
Register ati.is-a.dev

creeperkafasi:
Register ati.is-a.dev with CNAME record pointing to atiAI.github.io/ati.

### DIFF
--- a/domains/ati.json
+++ b/domains/ati.json
@@ -1,13 +1,11 @@
 {
-        "owner": {
-           "username": "creeperkafasi",
-           "email": "",
-           "discord": "442685489833705483",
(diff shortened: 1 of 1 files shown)


creeperkafasi:
I have a few things I want to clarify about this pr.

I already have a domain registered to my name, I want to register this one for an org I made for a hackathon me and some friends joined. I hope that's not a problem.

Also as you can see I used a path instead of the base domain. I'm not sure if this is possible so I apologize if it is not. In case it is not possible, can I have it point to just atiai.github.io?

DEV-DIBSTER (approved):

creeperkafasi:
Thanks for the quick response :pray: You guys are the best

PR #16841, opened 2024-10-27
Register reginaldfd.is-a.dev

RFelixD:
Register reginaldfd.is-a.dev with CNAME record pointing to reginaldfelixd.infinityfreeapp.com.

### DIFF
--- a/domains/reginaldfd.json
+++ b/domains/reginaldfd.json
@@ -1,0 +1,13 @@
+{
+        "owner": {
+           "username": "RFelixD",
+           "email": "",
+           "discord": "794364416904527932",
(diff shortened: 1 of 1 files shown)


iostpa (changes requested):
I don't think you have sufficient content on your website, since it seems to be both a template and a WIP. If your website isn't finished, finish that first and then get our domain.

RFelixD:
> I don't think you have sufficient content on your website, since it seems to be both a template and a WIP. If your website isn't finished, finish that first and then get our domain.

Okeay i'll finish it first
thanks for the information

PR #16861, opened 2024-10-28
Register fry69.is-a.dev

fry69:
Register fry69.is-a.dev with CNAME record pointing to fry69.dev.

### DIFF
--- a/domains/fry69.json
+++ b/domains/fry69.json
@@ -1,0 +1,13 @@
+{
+        "owner": {
+           "username": "fry69",
+           "email": "",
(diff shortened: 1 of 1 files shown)


iostpa (approved):
I guess its fine?

fry69:
> I guess its fine?

Thank you for merging, but Cloudflare does not like this CNAME setup to a proxy domain (my fault, I need to read more docs) -> `Error 1014`

Looks like I really need to point to the `workers.dev` URL with a CNAME, not a domain (at least that is that the docs say). PR incoming, sorry for this.

PR #16862, opened 2024-10-28
Register anaj00.is-a.dev

anaj00:
Register anaj00.is-a.dev with CNAME record pointing to anaj00.vercel.app.


iostpa (approved):
Seems good to me, though you might have to verify the domain via TXT, which you should go [here](https://is-a.dev/docs/vercel/)

PR #16893, opened 2024-10-30
Update almaz.json

AlmaziikDev:
## Requirements

- [+] The file is in the `domains` folder and is in the JSON format.
- [+] The file's name is all lowercased and alphanumeric. 
- [+] There is sufficient information at the `owner` field.


cupglassDEV:
No need the dots behind em

It is always covered by this dnsconfig https://github.com/is-a-dev/register/blob/9fa09f30406c9f525704a4bd3cb023ae677a0476/dnsconfig.js#L63

cupglassDEV:
If improvMX cant resolve it, we'll wait. Because is-a.dev has a temporary downtime yesterday, so the dns propagation may be slow

AlmaziikDev:
> If improvMX cant resolve it, we'll wait. Because is-a.dev has a temporary downtime yesterday, so the dns propagation may be slow

ah, okay then, thanks

cupglassDEV:
> > If improvMX cant resolve it, we'll wait. Because is-a.dev has a temporary downtime yesterday, so the dns propagation may be slow
> 
> ah, okay then, thanks

Its ok, i have trouble with that also. Even though my domain has a special Nameserver permission, but when i add a new record, it keeps failing like nothing

cupglassDEV:
oh no
image

iostpa (changes requested):

iostpa (approved):
Seems good to me now

PR #16894, opened 2024-10-30
Create and.json

EducatedSuddenBucket:
I need NS because I want to start a subdomain service and I promise NS records won't be misused for things like NSFW or dangerous things.


wdhdev:
Unfortunately we won't be allowing nested subdomaim services.

However we might end up offering these at some point (e.g. .org.is-a.dev)

PR #16963, opened 2024-11-02
Update jumanji.json

heyjumanji:
Added cname to existing domain

### DIFF
--- a/domains/jumanji.json
+++ b/domains/jumanji.json
@@ -6,6 +6,7 @@
   "record": {
     "A": ["172.66.47.44", "172.66.44.212"],
     "MX": ["mx1.improvmx.com", "mx2.improvmx.com"],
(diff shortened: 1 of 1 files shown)


domains/jumanji.json
```diff
@@ -6,6 +6,7 @@
   "record": {
     "A": ["172.66.47.44", "172.66.44.212"],
     "MX": ["mx1.improvmx.com", "mx2.improvmx.com"],
-    "TXT": "v=spf1 include:spf.improvmx.com ~all"
+    "TXT": "v=spf1 include:spf.improvmx.com ~all",
+    "CNAME": "heyjumanji.pages.dev"
```
Stef-00012, line 10: You can not use other records when you have a `CNAME` record

Either use `A` record with IPv4 or remove the other `A`, `TXT` and `MX` record

Stef-DP (changes requested):

PR #17038, opened 2024-11-07
Register Genl.is-a.dev

Imposterundetalelover:
Register Genl.is-a.dev with TXT record pointing to Genl Private.


cupglassDEV:
@Imposterundetalelover 
You name clearly tells alot about you

Now see the docs, this arent even valid 😭 https://is-a.dev/docs

PR #17067, opened 2024-11-09
Update yashasviallen.json

JavaProgswing:
Enable cloudflare proxy for my subdomain.


JavaProgswing:
Just to be clear, this is a ipv6 only site. It's accessible for ipv6 clients only. 
That's exactly why I needed the cloudflare proxy.

PR #16965, opened 2024-11-02
Create th.json

thaihwng:
## Requirements
All requirements listed must be met unless specified otherwise by a **maintainer**.
Please note that we reserve the rights not to accept any domain at our own discretion.

- [ ] The file is in the `domains` folder and is in the JSON format.
- [ ] The file's name is all lowercased and alphanumeric. (th.json)
- [ ] You have completed your website. 
- [ ] The website is reachable.  
- [ ] The CNAME record doesn't contain `https://` or `/`.  
- [ ] There is sufficient information at the `owner` field.  
## Additional Information
I am using Cloudflare's NS records to enable custom domain configurations, specifically for setting up email services associated with my domain. This setup allows for better management of DNS settings and improved email handling capabilities, such as enhanced security, deliverability, and flexibility in configuring mail records.

## Website Link/Preview
thaihwng.github.io/h

PR #16758, opened 2024-10-23
Update undefined.json

miteigidesu:
This pull request, is to re-register my domain for NS record. Here is my reason:
- I have a Raspberry Pi at home, and I want to try opening some kind of server like web server or Minecraft server to explore new things (or to study?). I need a domain with NS record to hide my public home IP behind Cloudflare Proxy, to be safe from bad guys.
- Sometimes I have school/school club projects that need a domain instead of my VPS public IP, you know, I just can't purchase a new domain just for a 1-month project. I don't want to make a pull request everytime I need one.
- It is much easier for me to manage my domain DNS via Cloudflare Dashboard.

I confirm I will not use the domain for any illegal/inappropriate/NSFW activities, nor give access to another person. If this happens, my domain will be suspended.

This is NOT AI-generated. I can't keep myself from professional writing. I read @cupglassDEV 's pull request to know how to add NS record, thanks :D

*(Accidently remove the Requirements template part below, hope this is fine :P)*

PR #17071, opened 2024-11-09
Create sctech.json

sctech-tr:
## Here's why I have NS records:
- I have a bunch of (really a bunch of) subdomains, so I can't keep track of JSON files.
- It's easier of me to manage those subdomains from Cloudflare.
- Freedom. Those NS records are there for people to **use** and there's nothing that can stop me from using them.
## Requirements

## Website Link/Preview
https://sctech.netlify.app

### DIFF
--- a/domains/sctech.json
+++ b/domains/sctech.json
@@ -1,11 +1,19 @@
 {
   "owner": {
-    "username": "sctech-tr"
+    "username": "sctech-tr",
+    "email": "gamerselimiko@gmail.com"
   },
   "record": {
     "NS": [
       "kyrie.ns.cloudflare.com",
       "susan.ns.cloudflare.com"
+    ],
+    "MX": [
+      "mx1.improvmx.com",
+      "mx2.improvmx.com"
+    ],
+    "TXT": [
+      "v=spf1 include:spf.improvmx.com ~all"
     ]
   }
 }
--- a/domains/sctech.json
+++ b/domains/sctech.json
@@ -1,0 +1,11 @@
+{
+  "owner": {
+    "username": "sctech-tr"
+  },
+  "record": {
+    "NS": [
+      "kyrie.ns.cloudflare.com",
+      "susan.ns.cloudflare.com"
+    ]
+  }
+}


cupglassDEV:
god, i need to forward my message again 😭

cupglassDEV:
Forwarded from #16965 with my [own comment](https://github.com/is-a-dev/register/pull/16965#issuecomment-2453354270)
(Why is it forwarded? Because the OG Pull Request do exactly like this, with the same exact reason)

If you want to know what, and whos going to get the NS permission:
1. Services that forcefully needed you to use NS for custom domains, like Aternos and playit.gg. Some, like cloudflare transform and cloudflare tunnel needed you to use a special NS with your is-a.dev subdomain (because is-a.dev already listed on the Public Suffix List, so it treats *is-a.dev*, *co.kr*, *net.id* like the root of it)
OR
2. Some unsupported is-a.dev records that you desperately need it for your homeserver, like this https://github.com/is-a-dev/register/pull/16758

We don't accept responses like "oh, i want NS because it's easy to administer, than having to make a PR on Github", NO

We already made the https://registration.is-a.dev so the process can be a little bit easier, and it fails. Because some of y'all are sending invalid request to register the domain. Now you need to register using the manual one

cupglassDEV:
It doesn't matter if you are doing it for freedom, too many subdomains, or such

sctech-tr:
but “too many subdomains” is a valid reason for using ns records in the docs

sctech-tr:
and why do you don’t accept ns records? they are just like any other record type, there is NO reason to prevent users from using them

CuteDog5695:
> but “too many subdomains” is a valid reason for using ns records in the docs, as in

No it ain't, like we have on the FAQ (?), over 3000 domains have to edit the file via github PRs, why can't you?

> and why do you don’t accept ns records? they are just like any other record type, there is NO reason to prevent users from using them

Can easily be abused since you can make records without us knowing (NS records are basically full control over the subdomain), that is why we restrict it to those who have a valid reason for them

iostpa (changes requested):
Please also state the reason why you want NS records

CuteDog5695:
Oh, just saw the reason. No, you did not provide a valid reason.

cupglassDEV:
> and why do you don’t accept ns records? they are just like any other record type, there is NO reason to prevent users from using them
@sctech-tr 
SEE, I TOLD YAA 😂

Now the maintainer comes in, i already **warned** you from the start. But you still keep insisting

cupglassDEV:
As always, read the fricking docs. If you tell me about "Oh i want NS bcs it exist like other records", that means it equals "Oh i want PTR because it exist like other records"

Diffrent usage, same danger

https://is-a.dev/docs/domain-structure

cupglassDEV:
If you want to use your free will, buy a domain. We really have limited support of records

wdhdev:
Sigh

PR #17077, opened 2024-11-09
Register utsabadhikari.is-a.dev

utsab1231:
Register utsabadhikari.is-a.dev with CNAME record pointing to personportfolio-e25d4.web.app.


iostpa (approved):
Seems good to me.

Issue #17103, opened 2024-11-10
Prettier workflow


wdhdev:
### What do you want added?
I want to implement a GitHub workflow that runs everytime a commit is pushed to the main branch, which runs Prettier:

```
prettier ./domains -w --tab-width 2 --print-width 120
```

jadifydotco:
Just use the api: https://prettier.io/docs/en/api.html 

I know very much help

cupglassDEV:
@jadifydotco 
> Just use the api: https://prettier.io/docs/en/api.html
> 
> I know very much help
You can't just do that, make a pr to help him also

He KNOWS that the docs exist

cupglassDEV:
I already made the PR for this, see #17105

PR #17105, opened 2024-11-10
Add prettier workflow according to #17103

ghost:
#17103


cupglassDEV:
Heres the docs on the cli command heheheh

https://prettier.io/docs/en/cli.html

cupglassDEV:
@wdhdev ill need to check first, in case of smth 😁

wdhdev (changes requested):
Are you able to make it commit using the @is-a-dev-bot account, using the `BOT` environment variable (a GitHub PAT token)?

It is likely that using the default actions user will fail due to branch protections.

cupglassDEV:
> Are you able to make it commit using the @is-a-dev-bot account, using the `BOT` environment variable?
> 
> It is likely that using the default actions user will fail due to branch protections.

Hmm, maybe using this, but point it to `register`

image

cupglassDEV:
tbh this is one, and only solution i found

cupglassDEV:
Updated

wdhdev (dismissed):

wdhdev (approved):

cupglassDEV:
yey

PR #17280, opened 2024-11-20
Updating trix.is-a.dev

tr1xem:
Satified Was switching account and repo

### DIFF
--- a/domains/_github-pages-challenge-tr1xem.trix.json
+++ b/domains/_github-pages-challenge-tr1xem.trix.json
@@ -1,0 +1,9 @@
+{
+  "owner": {
+    "username": "tr1xem",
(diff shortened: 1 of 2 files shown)


domains/_github-pages-challenge-tr1xem.trix.json
DEV-DIBSTER, line 1: So you just want to keep the verification but not the actual record (trix.json)?

DEV-DIBSTER (changes requested):

domains/_github-pages-challenge-tr1xem.trix.json
tr1xem: Yes After verification i could use it to any repo i want(which in this case will be website repo)

PR #17292, opened 2024-11-21
[NS Records] ploszukiwacz.json

ploszukiwacz:
I need ns records becouse:

I need the CNAME record with MX and TX not possible with is-a.dev but possible with cf
And I have a rpi4 running searxng at home, now i want to expose it to my is-a.dev domain. Now doing this without cf would mean exposing my home ip, and basically doxxing myself. Now with cf I can proxy it and not expose my home ip


wdhdev:
Will be fixed soon, pause this request for now.

PlOszukiwaczDEV:
Sadge

PlOszukiwaczDEV:
I will pause later on pc i can't on mobile

wdhdev:
Fixed this issue in #17369, please change this to a CNAME, and then enable proxying.

PR #17446, opened 2024-11-30
Register 7z.is-a.dev

XikiZxGS:
Register 7z.is-a.dev with CNAME record pointing to xikizxgs.github.io.


iostpa (approved):
I guess its fine

PR #17505, opened 2024-12-01
Update sono.json

getSono:
# Reason for NS records

- I have a Raspberry Pi at home and I sometime use it for a small cloud etc. 
 To protect my IP, I wanna set up NS records.

- When I just wanna test out something, then I don't wanna create a json file and wait till my domain gets accepted.

I hope that my PR gets approved and when not it's fine too.

> My english is not the best


Stef-DP (approved):
first one seems a good reason to me

PR #17653, opened 2024-12-09
Register redolent.is-a.dev

RedolentHalo:
Register redolent.is-a.dev with TXT record pointing to none.


iostpa (changes requested):
Both files are useless since they don't point to anything. Please make a new pull request that actually has valid records.

If you need help with what records we support, please consider reading [our documentation.](https://is-a.dev/docs)

PR #17676, opened 2024-12-11
Update sabbir.json

s8rr:
Hi I'm sabbir, from BANGLADESH. This pull request is to add NS to sabbir.is-a.dev domain.

I repurposed my old PC to a linux server. now i use that to experiment my projects. like i installed coolify on it and Pterodactyl etc. but making PR to add a domain to my project is to slow and confusing. The most important part is i use cloudflare Proxy to mask my ip. i have everything setup in my clouldflare dashboard with my old domain its about to expire and i don't have the financial power to renew it.

### DIFF
--- a/domains/sabbir.json
+++ b/domains/sabbir.json
@@ -4,6 +4,6 @@
     "email": "sabbir.apbn.edu@gmail.com"
   },
   "record": {
-    "CNAME": "sabbir1.pages.dev"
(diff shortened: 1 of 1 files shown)


s8rr:
i use cloudflare tunnel a lot and its will be easy for me to manage via cloudflare dashboard. and my domain is about to expire and i don't have any other domain so it will be very helpful.

PR #17685, opened 2024-12-12
Register terraegg.is-a.dev

TerraEgg:
- I have done all of the required things 

## Website Link/Preview
[terraegg.rf.gd](https://terraegg.rf.gd)

### DIFF
--- a/domains/terraegg.json
+++ b/domains/terraegg.json
@@ -1,0 +1,10 @@
+{
+  "description": "Portfolio for TerraEgg",
+  "owner": {
+    "username": "TerraEgg",
+    "email": "albumanyt@gmail.com"
+  },
+  "record": {
+    "CNAME": "ns1.byet.org"
+  }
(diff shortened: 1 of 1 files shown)


domains/terraegg.json
```diff
@@ -0,0 +1,10 @@
+{
+  "description": "Portfolio for TerraEgg",
+  "owner": {
+    "username": "TerraEgg",
+    "email": "albumanyt@gmail.com"
+  },
+  "record": {
+    "CNAME": "ns1.byet.org"
```
Stef-00012, line 8: for ns you need to  use NS records and have a valid reason to require them

Stef-DP (changes requested):

domains/terraegg.json
TerraEgg: Alright, I'll work on it thanks 👍

PR #17786, opened 2024-12-20
Register 1tfr.is-a.dev

gimmywashere:
Register footjob.is-a.dev with CNAME record pointing to ammo.lol.


domains/footjob.json
Stef-00012, line 1: footjob is nsfw so u r not getting this domain

Stef-DP (changes requested):

Stef-DP (approved):

PR #17883, opened 2024-12-28
Register info-svo.is-a.dev

TheFirstPy:
Register info-svo.is-a.dev with CNAME record pointing to aldar0809.github.io.


wdhdev (changes requested):
This does not seem to be a very developer related website.

TheFirstPy (commented):
Ok

PR #17968, opened 2025-01-02
Create benrogo.json

BenRogozinski:
Added site json file

## Requirements

## Reason for NS records
I am going to be using this domain with Cloudflare tunnels, which requires you to have the domain in your Cloudflare account, so it's necessary to use Cloudflare's name servers for my project. My project is a service that allows people to embed proxied sites into their own website (note that this does not allow them to modify the contents of those proxied sites). For example, someone could embed www.google.com in their site but it would be proxied through benrogo.is-a.dev. This is especially useful for sites that would need to provide a preview of a website in its current state. Cloudflare tunnels are being used to load balance between multiple servers and have the service automatically choose the one that is geographically the closest. An example of something similar (owned my friend technonyte) is hosted at [https://alkali.pages.dev/static/](https://alkali.pages.dev/static/). Another reason for needing NS records is to more easily make subdomains of benrogo.is-a.dev, in order to allow communities to have their own subdomains.

### DIFF
--- a/domains/benrogo.json
+++ b/domains/benrogo.json
@@ -1,0 +1,10 @@
+{
+  "owner": {
+    "username": "BenDaBuilder",
+    "discord": "Ben_Da_Builder",
+    "email": "brogodev@gmail.com"
+  },
+  "record": {
+    "NS": ["eve.ns.cloudflare.com", "john.ns.cloudflare.com"]
+  }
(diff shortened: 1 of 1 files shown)


wdhdev:
I'm going to say no as this could be a security and abuse risk.

PR #17979, opened 2025-01-02
lowercase

Stef-DP:
who cares

### DIFF
--- a/tests/pr.test.js
+++ b/tests/pr.test.js
@@ -2,7 +2,7 @@ const t = require("ava");
 const fs = require("fs-extra");
 const path = require("path");
 
-const PR_AUTHOR = process.env.PR_AUTHOR;
+const PR_AUTHOR = process.env.PR_AUTHOR.toLowerCase();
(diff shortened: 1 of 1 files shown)


wdhdev:
You also need to update the vars that reference the actual file owner

wdhdev:
Can you make it so the admins array will filter to find the user in lowercase

wdhdev (dismissed):

wdhdev:
More issues, it fails if you're updating something that isn't a file

Stef-00012:
> More issues, it fails if you're updating something that isn't a file

I noticed that the action file checks for changed files only in the `domains/` folder
```yml
- name: Get all modified domain JSON files
  if: github.event_name == 'pull_request'
  id: changed-files
  uses: tj-actions/changed-files@v45
  with:
      files: domains/*.json
```

so i renamed the variable from `MODIFIED_FILES` to `MODIFIED_DOMAIN_FILES` and made it do it becomes `null` if the env variable `process.env.MODIFIED_FILES` is empty (length is not `> 0`), then in the checks i just made it so it passes if there's no `MODIFIED_DOMAIN_FILES`

is this ok?

@wdhdev

wdhdev (approved):

PR #18258, opened 2025-01-11
Create manugeni.json

xrendani:


domains/manugeni.json
```diff
@@ -0,0 +1,8 @@
+{
+  "owner": {
+    "username": "manugeni"
+  },
+  "record": {
+    "CNAME": "manugeni.github.io"
```
DEV-DIBSTER, line 6: Uncompleted site.

DEV-DIBSTER (changes requested):

domains/manugeni.json
manugeni: Could you check again.

PR #18429, opened 2025-01-14
Create _github-pages-challenge-corebell.corebell.json

xrendani:


domains/_github-pages-challenge-corebell.corebell.json
Stef-00012, line 1: you can not create a subdomain of `corebell.is-a.dev` as it does not exist

Stef-DP (changes requested):

PR #18492, opened 2025-01-15
Hill climb racing 

EducatedSuddenBucket:
Hill climb racing

### DIFF
--- a/domains/educatedsuddenbucket.json
+++ b/domains/educatedsuddenbucket.json
@@ -5,10 +5,7 @@
   },
   "record": {
     "A": [
-      "185.199.108.153",
-      "185.199.109.153",
-      "185.199.110.153",
-      "185.199.111.153"
+      "185.199.108.153"
     ],
     "TXT": [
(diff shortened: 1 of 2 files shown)


wdhdev (changes requested):
Why are you removing the other GitHub Pages IPs? You need those.

EducatedSuddenBucket:
@wdhdev only one ip is enough u can check that https://random.esb.is-a.dev works and it has only one record ,https://digger.tools/lookup/random.esb.is-a.dev

EducatedSuddenBucket:
@wdhdev only one ip is enough u can check that https://random.esb.is-a.dev works and it has only one record ,https://digger.tools/lookup/random.esb.is-a.dev

wdhdev:
I mean yes. However for redundancy you shpuld have all 4.

EducatedSuddenBucket:
I just wanna keep it simple with one IP. GitHub Pages already handles redundancy on their end, so extra IPs don't really matter for me.

wdhdev (approved):
I mean sure. Don't see how it causes an issue having all 4 either. Also, I'd assume the redundancy is through having multiple IPs.

PR #18702, opened 2025-01-20
Register seranok.is-a.dev

seranok1337:
Register seranok.is-a.dev with CNAME record pointing to seranok1337.github.io.


domains/seranok-is-a.dev
Stef-00012, line 1: delete this file

Stef-DP (changes requested):

PR #19798, opened 2025-02-10
Register chat.is-a.dev

wonder1418:
Register chat.is-a.dev with CNAME record pointing to 093357dd-a6e7-4693-94a7-0520951ca954.cfargotunnel.com.


wdhdev:
Too generic of a name.

PR #20424, opened 2025-02-22
fix(ci): validation issue for CNAMEs

DEV-DIBSTER:
From PR: https://github.com/is-a-dev/register/pull/20396

CNAME validation somehow passed with HTTPS present within the domain. It should also address file paths.


wdhdev:
This PR would be rendered null as the validation only runs on JSON files, in the PR you linked, it had a `.js` extension and it was in the root directory, so that's why tests passed, as it is technically a valid file as it was a JavaScript file.

PR #20457, opened 2025-02-22
[not subdomain] suggest some subdomains to be reserved

justdevw:
```json
"cname",
"cname[1-99]",
"srv",
"srv[1-99]",
"txt",
"txt[1-9999]"
```


Stef-00012:
wonder how dumb someone would need to be to use a record as domain name

DevMatei:
> wonder how dumb someone would need to be to use a record as domain name

Everything is quite possible in our world 😭

wdhdev (approved):

PR #20588, opened 2025-02-26
chore(domain): add missing A record

JonathanXDR:
# Requirements

- [x] I have **read** and **understood** the [Terms of Service](https://is-a.dev/terms). 
- [x] I understand my domain will be removed if I violate the [Terms of Service](https://is-a.dev/terms).
- [x] My file is in the `domains` directory and has the `.json` file extension.
- [x] My file's name is lowercased and alphanumeric. 
- [x] My website is **reachable** and **completed**. 
- [x] I have provided sufficient contact information in the `owner` key. 

# Website Preview

[jonathan-russ.com](https://jonathan-russ.com/)

### DIFF
--- a/domains/_vercel.jonathan-russ.json
+++ b/domains/_vercel.jonathan-russ.json
@@ -5,6 +5,7 @@
     "discord": "494042822320717827"
   },
   "record": {
-    "TXT": "vc-domain-verify=jonathan-russ.is-a.dev,298a207b5f94dffde572"
+    "TXT": "vc-domain-verify=jonathan-russ.is-a.dev,298a207b5f94dffde572",
+    "NS": ["ns1.vercel-dns.com", "ns2.vercel-dns.com"]
   }
 }
--- a/domains/jonathan-russ.json
+++ b/domains/jonathan-russ.json
@@ -7,6 +7,6 @@
     "discord": "494042822320717827"
   },
   "record": {
-    "NS": ["ns1.vercel-dns.com", "ns2.vercel-dns.com"]
+    "A": ["76.76.21.21"]
   }
 }


JonathanXDR:
The tests fail due to the A record. The error states: `NS records cannot be combined with other records, except for DS records.` However, this combination seems necessary for the connection to Vercel through the `_vercel.subdomain.json` file, which already exists for my subdomain. Whats wrong here?

domains/_vercel.jonathan-russ.json
```diff
@@ -5,6 +5,7 @@
     "discord": "494042822320717827"
   },
   "record": {
-    "TXT": "vc-domain-verify=jonathan-russ.is-a.dev,298a207b5f94dffde572"
+    "TXT": "vc-domain-verify=jonathan-russ.is-a.dev,298a207b5f94dffde572",
+    "NS": ["ns1.vercel-dns.com", "ns2.vercel-dns.com"]
```
21Z, line 9: you cant use another record with NS records

21Z (changes requested):

domains/_vercel.jonathan-russ.json
JonathanXDR: Hmm okay, but how can I provide a TXT for vercels verification und then set the nameservers if that's not possible?

iostpa: If Vercel DNS is being a pain in the ass, switch to Cloudflare DNS.

JonathanXDR: I configured another domain (app.is-cool.dev) from this repo: https://github.com/open-domains/register, and as you can see, Vercel is expecting either an A record or the nameservers (my PR just hasn't been approved for that repo yet, which is why it's not connected yet).
However, with this project, Vercel only expects a TXT record, which doesn't work for my case as I want to set the nameservers. Could it be that you as maintainers need to add some extra configuration in Cloudflare or something for Vercel to recognize that?
image

iostpa: I think its how Vercel added us to the list.
Back when we asked them to add us so that is-a.dev users could use Vercel they made it like that (check [this issue](https://vercel.community/t/is-a-dev-domain-issue/923) that we made to get us added), but now I'm not sure if that is the only way you can use Vercel, I could be wrong really.

For simplicity, you can consider using another DNS provider like Cloudflare.

PR #21083, opened 2025-03-11
Delete domains/vai.devmatei.json

21Z:
[for maintainers](https://discord.com/channels/830872854677422150/830872946888146964/1348337190597296139)

### DIFF
--- a/domains/vai.devmatei.json
+++ b/domains/vai.devmatei.json
@@ -1,9 +1,0 @@
-{
-    "owner": {
-      "username": "DevMatei",
-      "email": "contact@devmatei.is-a.dev"
-    },
-    "record": {
-      "A": ["75.2.60.5"]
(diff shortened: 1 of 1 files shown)


JustDeveloper1:
what the heck went wrong with maintainers channel description

JustDeveloper1:
image

JustDeveloper1:
btw im not using plugins, i clicked that link 
image

iostpa (approved):
Approving since the domain is not available atm (not important), but its also being used for blackmail in my opinion.

This should be reported to is-a.dev maintainers if the behavior continues.

iostpa:
> This should be reported to is-a.dev maintainers if the behavior continues.

By behaviour I mean Vai's behavior, if that makes sense.

PR #21500, opened 2025-03-25
Remove domains breaking ToS

21Z:
website is not developer related at all and is just doxxing.


wdhdev:
All his domains should be removed in this case

wdhdev (dismissed):

wdhdev (approved):

PR #21528, opened 2025-03-26
Register kembabj.is-a.dev with Cloudflare NS

kakakakzz:
I’m requesting NS records for `kembabj.is-a.dev` to use Cloudflare nameservers. I’ve updated my reasoning based on feedback to clarify my needs:

- My project "Key Management" is a personal project to manage activation keys for tools and software. It is deployed on a VPS and currently operates via `http://ip:port`, exposing the API key URL over HTTP. When I integrate this API into a Python tool and share the tool, the API key could be exploited by malicious actors. Therefore, I need a domain to secure the API key, and I chose `is-a.dev` because of its appealing name and free service.

- I need Cloudflare NS to easily manage DNS records through Cloudflare’s dashboard, as well as to leverage its free firewall and rate limiting features to protect my project from potential attacks (like DDoS) since my VPS uses a static IP address. Cloudflare NS also enables HTTPS, encrypting the connection and protecting the API key URL.

- I commit to using this subdomain for legitimate personal purposes and will not abuse it to harm the reputation of is-a.dev. I understand the concerns about abuse, and I’m willing to cooperate if the is-a.dev team needs further verification (via email or Discord). My project is legitimate.


kakakakzz (commented):
sorry i forgot to add .json

iostpa:
Just to add a note, only the owner of this service can merge domains related to NS records, thats why I added the label.

tranchikhanh:
Thanks for the advice @MaskDuck

PR #21903, opened 2025-04-08
Chage NS records from cloudflare to own hosted

lungustefan:
# N/A


lungustefan:
@wdhdev / @DEV-DIBSTER The json syntax is valid. The check ain't - It is dumb trust me check json syntax

iostpa:
~~blame willy~~
To be fair I don't think he expected people to add multiple DS records on a single domain, but it still needs to be fixed anyways.

Stef-DP (dismissed):

lungustefan:
> t think he expected people to add multiple DS records on a single domain, but it still needs to be fixed anyways.

Yes ciz no DNS provider asks for more than one DS yk

Stef-00012:
let's be honest, willy sucks at coding so he made bad checks fr fr

lungustefan:
> let's be honest, willy sucks at coding so he made bad checks fr fr

Let's be honest! This should go in PINS in general

DEV-DIBSTER:
Appears to be with the way the `/tests/json.test.js` at [findDuplicateKeys](https://github.com/is-a-dev/register/blob/main/tests/json.test.js#L38-L53) handles duplicates to see if they're the same.

```js
function findDuplicateKeys(jsonString) {
    const keyPattern = /"([^"]+)"(?=\s*:)/g;
    const keys = [];
    let match;

    while ((match = keyPattern.exec(jsonString)) !== null) {
        keys.push(match[1]);
    }

    const keyCount = {};
    keys.forEach((key) => {
        keyCount[key] = (keyCount[key] || 0) + 1;
    });

    return Object.keys(keyCount).filter((key) => keyCount[key] > 1);
}
```

This regex will only check the very basics of it, and doesn't look into properties of the DS records.

lungustefan:
> Appears to be with the way the `/tests/json.test.js` at [findDuplicateKeys](https://github.com/is-a-dev/register/blob/main/tests/json.test.js#L38-L53) handles duplicates to see if they're the same.
> 
> ```js
> function findDuplicateKeys(jsonString) {
>     const keyPattern = /"([^"]+)"(?=\s*:)/g;
>     const keys = [];
>     let match;
> 
>     while ((match = keyPattern.exec(jsonString)) !== null) {
>         keys.push(match[1]);
>     }
> 
>     const keyCount = {};
>     keys.forEach((key) => {
>         keyCount[key] = (keyCount[key] || 0) + 1;
>     });
> 
>     return Object.keys(keyCount).filter((key) => keyCount[key] > 1);
> }
> ```
> 
> This regex will only check the very basics of it, and doesn't look into properties of the DS records.

Yep but a force merge could and should work as of your handling for putting the ns, but better to wait for William to check

wdhdev (approved):

PR #22417, opened 2025-04-28
added upchat.is-a.dev

vk0900:
added upchat.is-a.dev

### DIFF
--- a/domains/upchat.json
+++ b/domains/upchat.json
@@ -1,0 +1,10 @@
+{
+  "description": "chat website",
+  "owner": {
+    "username": "VK",
+    "email": "nnmvishnu@gmail.com"
+  },
+  "records": {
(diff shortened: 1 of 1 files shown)


Stef-DP (changes requested):
you need to provide a valid reason to use NS records

vk0900:
i want to use my own ns records so i can manage DNS settings from Cloudflare and i get  DDoS protection and automatic SSL/TLS and where else do you get to run ts for free and this is my first website it's a chat website for low latency which Cloudflare helps

vk0900:
why am i using a NS records because i what to use cloudflare to host my website and i can manage DNS settings from Cloudflare and i get DDoS protection and automatic SSL/TLS and where else do you get all of this for free and this is my first website it's a chat website for low latency which Cloudflare helps

PR #22676, opened 2025-05-10
Create url-encryption-tool.json

tijohncenna:
I have Url Encryption Tool, fully made of JavaScript hosted on Cloudflare Workers. Now I want to use a custom domain with "is-a.dev". So I have to add a route in the workers dashboard and to do that I must have to add a zone/domain firstly; that's why I requested for NS method. I attached a screenshot of the issue.

IMG_20250510_234529

### DIFF
--- a/domains/url-encryption-tool.json
+++ b/domains/url-encryption-tool.json
@@ -1,0 +1,14 @@
+{
+  "owner": {
+    "username": "tijohncenna",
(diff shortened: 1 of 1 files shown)


wdhdev:
I'm going to deny this as non-dev related. is-a.dev subdomains are mainly only to be used for portfolios and such (well the root domains are.)

PR #23332, opened 2025-06-04
chore: adding secure to reserved subdomains

DEV-DIBSTER:
Adding two additional subdomains to reserved list.


util/reserved.json
```diff
@@ -27,6 +28,7 @@
     "doc",
     "documentation",
     "domain",
+    "donate",
```
wdhdev: Exists https://github.com/is-a-dev/register/blob/main/domains/donate.json

util/reserved.json
```diff
@@ -3,6 +3,7 @@
     "account",
     "accounts",
     "acme",
+    "access",
```
wdhdev: Please alphabetise

wdhdev (changes requested):

wdhdev (approved):

PR #24285, opened 2025-07-10
[not a subdomain] add official_subdomains.json to raw api

justdevw:
https://discord.com/channels/830872854677422150/1341746804823162941/1391062147412525167


wdhdev (changes requested):
This makes zero sense when we have internal domains

PR #24303, opened 2025-07-10
Create cityguesser.json

TizianGth:


dragsbruh:
sorry we do not allow non dev related sites on root subdomain

PR #24842, opened 2025-07-30
Add a DNS TXT record

RzkyF:
# Requirements

- [X] I have **read** and **agree** to the [Terms of Service](https://is-a.dev/terms). 
- [X] My file is following the [domain structure](https://docs.is-a.dev/domain-structure/).
- [X] My website is **reachable** and **completed**. 
- [X] My website is **software development** related. 
- [X] My website is not for commercial use. 
- [X] I have provided sufficient contact information in the `owner` key. 
- [X] I have provided a preview of my website below. 

# Website Preview

Here is my portofolio pages : https://rzkyf.github.io/

PR #24843, opened 2025-07-30
Revert "Add a DNS TXT record"

iostpa:
Reverts #24842 due to my stupidity.


iostpa:
image
As to why I'm removing it

PR #25013, opened 2025-08-07
add ljx.json

turkey1024:
# Requirements

The reason why I added NS records is that I am in China and need CloudFlare's CDN acceleration function, otherwise the speed will be as slow as a snail. I promise not to abuse the domain name.

- [x] I have **read** and **agree** to the [Terms of Service](https://is-a.dev/terms). 
- [x] My file is following the [domain structure](https://docs.is-a.dev/domain-structure/).
- [x] My website is **reachable** and **completed**. 
- [x] My website is **software development** related. 
- [x] My website is not for commercial use. 
- [x] I have provided sufficient contact information in the `owner` key. 
- [x] I have provided a preview of my website below. 

The reason why I added NS records is that I am in China and need CloudFlare's CDN acceleration function, otherwise the speed will be as slow as a snail. I promise not to abuse the domain name.

# Website Preview

### DIFF
--- a/domains/ljx.json
+++ b/domains/ljx.json
@@ -1,0 +1,16 @@
+{
+"owner"
+: {
+"username"
+: "turkey1024",
+"email"
+: "ljx20110119@qq.com"
+},
+"records"
+: {
+"NS"
+: ["daisy.ns.cloudflare.com", "hal.ns.cloudflare.com"]
+},
+"proxied"
+: true
+}


ijasxyz:
Your website preview is broken.

dragsbruh:
if you want the "cdn acceleration" feature you can always use the proxied feature we provide t hat does the exact same thing but without the NS record mess

also the format for your json file is a bit weird?

https://docs.is-a.dev/domain-structure/#-proxied-optional

PR #25218, opened 2025-08-14
Migration from clyron.is-a.dev to kay.is-a.dev

AveryEclipse:
This pull request:
- registers kay.is-a.dev, replacing https://clyron.is-a.dev which is the current website
     - basically kay.is-a.dev will be the same clyron.is-a.dev, using the same CNAME record (and thus the same website github repo) as before
- renames the ATproto and Discord verification files so they assure my website is kay.is-a.dev, NOT clyron.is-a.dev
- clyron.is-a.dev and its nested subdomains remain, but its CNAME records are changed to URL records pointing to the new equivalent kay.is-a.dev addresses to minimize link rot, *at least for the time being*
    - I suppose I have done this right, otherwise I am willing to take notes


kay.json
STICKnoLOGIC, line 1: move this inside of "domains" folder

tweets.kay.json
STICKnoLOGIC, line 1: move this inside the "domains" folder

STICKnoLOGIC (changes requested):

kay.json
AveryEclipse: fixed!
same for tweets.kay.json

i overlooked and thought they already were inside the domains folder, now they should be there

STICKnoLOGIC (approved):

iostpa (approved):
Seems good to me

PR #25249, opened 2025-08-14
Remove some hijacked domains

iostpa:
@nilsraccoon @korbirayen @SkSakilAli @theCode-Breaker @Dn1w Your domain's are being removed because the domains point to a different website. (Basically it got hijacked by someone)


korbirayen:
how did that even happen

iostpa:
> how did that even happen

GitHub Pages is pretty lenient on security, so unless the domain was verified or still had a website connected to it anyone could just take it and use it.

PR #25293, opened 2025-08-15
Update TXT record for kay.is-a.dev Discord verification

AveryEclipse:
I didn't expect to have to create another PR right after my last one but yeah uh turns out a different DH is needed if you want to verify a different domain with Discord
Mistakes happen

### DIFF
--- a/domains/_discord.kay.json
+++ b/domains/_discord.kay.json
@@ -4,6 +4,6 @@
     "email": "thekayaurora@gmail.com"
   },
   "records": {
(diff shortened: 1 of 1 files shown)


iostpa (approved):

thekayaurora:
> I think only one discord account can have the domain, not sure.
No, that's not the issue. The two DHs that were in the .json file before *were* for two separate discord accounts, but both were meant for a different is-a.dev subdomain (the one that was before kay.is-a.dev), so I'm assuming it's a subdomain issue

PR #28830, opened 2025-10-31
Add kaharu.is-a.dev

Kaharu7:
Added kaharu.is-a.dev
This domain will be used for my AI watermark remover and text-to-speech generator website.
IMG-20251031-WA0005


STICKnoLOGIC (commented):
closing:
- not dev related, means we need a portfolio/personal site for subdomain and all projects must be in nested subdomain

PR #29267, opened 2025-11-12
Create kinoplus4k.json

FoxStudio24:
Add kinoplus4k.is-a.dev


Stef-DP (changes requested):
root domain must be dev related and streaming services are not allowed, see https://is-a.dev/terms

PR #29638, opened 2025-11-21
added shxmeem.json

Shxmeem:
Adding Cloudflare name servers for shxmeem.json

I accept all the terms and conditions and i won't use for unethical purposes or commercial.
I'm using xz.is-a.dev so demo is there.
Screenshot_20251121_224311_Brave

i can upload here as well.


iostpa (changes requested):
You can't get NS records while registering for a domain, you need to own the domain for at least a month.

PR #29747, opened 2025-11-24
Add 0xio32.is-a.dev

0xIO32:
# Requirements

- [X] I **agree** to the [Terms of Service](https://is-a.dev/terms). 
- [X] My file is following the [domain structure](https://docs.is-a.dev/domain-structure/). 
- [X] My website is **reachable** and **completed**. 
- [X] My website is **software development** related. 
- [X] My website is **not for commercial use**. 
- [X] I have provided contact information in the `owner` key. 
- [X] I have provided a preview of my website below. 

# Website Preview

image
Link: https://0xio32-is-a-dev.pages.dev
<br><br>
Some notes to the website:<br>
- Mimics fish + starship + bat in a terminal (it uses xterm.js under the hood) <br>
- Some of the elements in the terminal are clickable

# Reasons for NS record

Preamble: If the reasons are not clear enough, just ask and I will provide more info.

> We allow NS records only when:
> The standard DNS record types we support aren't sufficient.

The "HTTPS" dns record seems nice to have (advertising HTTPS v3, Quic support. Encrypted ClientHello afters its implemented in browsers).

> A provider (e.g. Aternos) requires NS delegation and does not have any other method of setting up a subdomain.
> This is mostly regarding SaaS providers, not DNS providers like Cloudflare or deSEC.

1. This isn't a problem for the root website, but I want to delegate some subdomains to my NAS, which is under a "Dynamic IP", without an API (like the one Cloudflare provides) to update the ip, this wouldn't be possible.

2. Cloudflares email routing isn't available, if the domain isn't registered in Cloudflare.

> We do not allow NS records for:
> Self hosting
> Note: This can vary from case to case

I like to have everything under one subdomain, the dev website and some self-hosted services, if thats ok.

> Convenience or ease of use

I don't use it as reason, but Cloudflare would definitely provide it.

> Unverifiable or unsupported claims
> Non-functional or placeholder purposes
> Self hosting websites that can be easily hosted on a free platform like GitHub Pages

Not the case

> If we have security or abuse concerns

I hope thats not the case.

> New users of is-a.dev
> If you have not had an is-a.dev subdomain for at least 30 days, you will be denied due to security reasons.

I don't see the point, but if it makes you feel safer, I can add another pr in 30 days after the domain points to the cloudflare pages.

> Users that have abused is-a.dev subdomains in the past

Nothing that I am aware of

<br>

Thanks
\- Markus Probst

### DIFF
--- a/domains/0xio32.json
+++ b/domains/0xio32.json
@@ -1,0 +1,9 @@
+{
+  "owner": {
+    "username": "0xio32",
+    "email": "markus.probst@posteo.de"
+  },
+  "records": {
+    "CNAME": "0xio32-is-a-dev.pages.dev"
+  }
+}


iostpa (changes requested):
So here's the problem.
We only allow NS records on domains that have existed for a month after registration. Sorry but you have to use the domain for your website with records like A or CNAME.

0xIO32:
I suppose the NS record will be a christmas gift then? (25.11.2025 + 30 days -> 25.12.2025)

domains/0xio32.json
```diff
@@ -0,0 +1,9 @@
+{
+  "owner": {
+    "username": "0xio32",
+    "email": "markus.probst@posteo.de"
+  },
+  "records": {
+    "CNAME": "website.markustieger6851.workers.dev"
```
iostpa: You can't connect a domain directly with CF Workers, try CF Pages if you can.

iostpa (changes requested):

domains/0xio32.json
0xIO32: Its the same thing, at least on my dashboard.
image

An older domain seems to got and still has a .pages.dev domain though.

iostpa (approved):
Seems good to me now (you should probably fix the error when its the default zoom though)

PR #29872, opened 2025-11-27
0

GiorgioKhoury:


FWEEaaaa1:
yk you cant put ur json file in the root directory ;sob:
it should be in domains

PR #30287, opened 2025-12-09
Resolve issues

STICKnoLOGIC:
The domains violates is-a-dev ToS

Closes #30221


iostpa (approved):
For `ff.shaid.is-a.dev` I really don't think it should of been allowed since its selling stuff, kind of contradicting the description.

PR #30419, opened 2025-12-13
Add traderpulse.is-a.dev

m-sadeghian:
Personal use

### DIFF
--- a/domains/traderpulse.json
+++ b/domains/traderpulse.json
@@ -1,8 +1,10 @@
 {
+  "description": "Personal VPN server",
   "owner": {
-    "username": "m-sadeghian"
+    "username": "m-sadeghian",
   },
   "record": {
     "A": ["89.106.69.249"]
-  }
+  },
+  "proxy": false
 }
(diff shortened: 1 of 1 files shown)


domains/traderpulse.json
```diff
@@ -0,0 +1,10 @@
+{
+  "description": "Personal VPN server",
+  "owner": {
+    "username": "m-sadeghian",
+  },
+  "record": {
```
ok-coder1, line 6:
```suggestion
  "records": {
```
It should be `records` and not `record`.

domains/traderpulse.json
```diff
@@ -0,0 +1,10 @@
...
+    "A": ["89.106.69.249"]
```
ok-coder1, line 7: This site is not reachable.

domains/traderpulse.json
```diff
@@ -0,0 +1,10 @@
...
+  },
+  "proxy": false
```
ok-coder1, line 9:
```suggestion
```
If you don't want Cloudflare proxy, why even bother typing it in to set it false?
Also, `proxy` should be `proxied`.

ok-coder1:
Please also fill in the PR template.

STICKnoLOGIC (commented):
Closing
- probably commercial (due to trading)

PR #30675, opened 2025-12-21
Create sus.clearcraft.json

Minecraft1220:
I want to use it for WebSocket.


STICKnoLOGIC (commented):
Unauthorized, you cant create a nested subdomain if you dont get/own the root subdomain

PR #30722, opened 2025-12-22
Add TXT record for domain verification

Infoaabir:


STICKnoLOGIC (commented):
Closing:
not the way to add txt verification, you need to add nested subdomain like _vercel.xxx.json and add the TXT recods (which is an instance of STRING)

read more at step  -> https://docs.is-a.dev/guides/vercel/

PR #30787, opened 2025-12-24
Replace A record with TXT record for domain verification

Infoaabir:


STICKnoLOGIC (commented):
read the docs, specially the step 5 -> https://docs.is-a.dev/guides/vercel/

PR #31064, opened 2026-01-01
Update duckers.json

TheRealDuckers:
Just adding MX bits

### DIFF
--- a/domains/duckers.json
+++ b/domains/duckers.json
@@ -3,10 +3,10 @@
         "username": "TheRealDuckers"
     },
     "records": {
-        "CNAME": "therealduckers.github.io"
-        "MX": [
-  { "target": "feedback-smtp.eu-west-1.amazonses.com", "priority": 10 }
-]
+"CNAME": "therealduckers.github.io",
+"MX": [
+"feedback-smtp.eu-west-1.amazonses.com"
+],
         "TXT": "v=spf1 include:amazonses.com ~all"
     }
 }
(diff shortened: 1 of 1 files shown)


domains/duckers.json
FWEEaaaa1, line 1: u need to add a comma at the end of ur cname file, and a comma at the end of the MX array "]," like that and it will be fixed
my advice is run it through a json validator like https://jsonlint.com

domains/duckers.json
```diff
@@ -4,5 +4,9 @@
     },
     "records": {
         "CNAME": "therealduckers.github.io"
+        "MX": [
+  { "target": "feedback-smtp.eu-west-1.amazonses.com", "priority": 10 }
+]
```
STICKnoLOGIC: json key is separated by comma "," and the default priority of mx is 10
```suggestion
"CNAME": "therealduckers.github.io",
"MX": [
"feedback-smtp.eu-west-1.amazonses.com"
],
```

domains/duckers.json
```diff
@@ -4,5 +4,9 @@
...
+        "TXT": "v=spf1 include:amazonses.com ~all"
     }
```
STICKnoLOGIC: you cant combine/mix CNAME with other records unless proxies
```suggestion
    },
    "proxied": true
```

STICKnoLOGIC (changes requested):
- resolve issues below

STICKnoLOGIC (approved):

PR #31127, opened 2026-01-03
Create prehistoricvtt.is-a.dev

dinosaurscience078-ux:
Requesting prehistoricvtt.is-a.dev domain


iostpa (changes requested):
You should of left the pull request template as-is and completed it, file format is invalid since "record" should be "records" and you can't add a is-a.dev subdomain to Cloudflare Tunnels directly.

PR #31381, opened 2026-01-09
Create _dmarc.gupta.json

Priyansxu:


iostpa (approved):
@STICKnoLOGIC This is fine, he has a different domain for the website but wants to use a different domain for the email.

PR #32683, opened 2026-02-11
Register project-aiman.is-a.dev

CZJindrich:
PR closed — will resubmit with corrections.


LonelyGuy12:
You can't use a GitHub no reply email
The error in the CI is

```log
  project-aiman.json: Owner email should not be a GitHub no-reply email
```

PR #34854, opened 2026-03-19
add tweetly.is-a.dev

0x0Megasus:
Adding tweetly.is-a.dev for my social app Tweetly.


STICKnoLOGIC (commented):
Closing;
- not dev related, means we need a "software dev" related portfolio/```personal``` site for subdomain and all ```projects``` must be in nested subdomain
- removing the entire PR template will result to auto close

PR #35029, opened 2026-03-22
add kamis5.is-a.dev

Jajerito95:
Adding my personal portfolio domain kamis5.is-a.dev pointing to mi-portafolio-seven-iota.vercel.app


iostpa (changes requested):
The screenshare guide raises some questions for me, since its common for scammers to use it. Why do you have a guide to do what scammers usually do to naive people just to catch if someone is not hacking in Minecraft?

Also do you accept our Terms of Service?

PR #35069, opened 2026-03-23
Register mindprint.is-a.dev

hamzabenkhdher-byte:
Registering my domain for a Cloudflare Tunnel project.

### DIFF
--- a/domains/mindprint.json
+++ b/domains/mindprint.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "your-github-username",
+        "email": "your-email@example.com"
+    },
+    "record": {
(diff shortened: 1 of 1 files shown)


domains/mindprint.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "your-github-username",
+        "email": "your-email@example.com"
+    },
+    "record": {
+        "CNAME": "f3b045a9-ec43-4a5a-87bb-800e71b26bc1.cfargotunnel.com"
```
dragsbruh, line 7: thats not how you use cloudflare tunnels
+ cloudflare tunnels need ns records, which you are not eligible for
+ all main domains here are supposed to be developer related

see [docs](https://docs.is-a.dev/) or [README](https://github.com/is-a-dev/register/blob/main/README.md) for more info

PR #35129, opened 2026-03-24
Report abuse

kenjiew:
domain
https://anji.is-a.dev
and
https://projects.itsmarian.is-a.dev
tag
@itsmarianmc 
and
@ianjiteshan 

It looks like your domain has been hijacked


satr14washere:
one gambling and one is offline

PR #35174, opened 2026-03-25
feat: add lev.is-a.dev

karamanlilevent:
Registering lev.is-a.dev for Claude PM dashboard

### DIFF
--- a/domains/lev.json
+++ b/domains/lev.json
@@ -1,0 +1,9 @@
+{
+  "owner": {
+    "username": "karamanlilevent",
+    "email": "karamanlilevent@gmail.com"
+  },
+  "record": {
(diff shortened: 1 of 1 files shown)


domains/lev.json
```diff
@@ -0,0 +1,9 @@
+{
+  "owner": {
+    "username": "karamanlilevent",
+    "email": "karamanlilevent@gmail.com"
+  },
+  "record": {
```
justdevw, line 6:
```suggestion
  "records": {
```

iostpa (changes requested):
No website preview, though judging by the name it could be something which we do not allow. Invalid file as well.

PR #35227, opened 2026-03-26
add digital-leaner.is-a.dev

ms7s917000-cpu:
Registering digital-leaner.is-a.dev pointing to ms7s917000-cpu.github.io


iostpa (changes requested):
Website preview doesn't work but judging by the name I don't think we would allow it.

PR #35256, opened 2026-03-27
Add dwh.is-a.dev

sanchez1987-1:
## Domain Registration
- Subdomain: dwh.is-a.dev
- IP: 95.174.94.6
- Owner: sanchez1987-1


domains/server.json
justdevw, line 1: `server.is-a.dev` is reserved

PR #35316, opened 2026-03-28
Report Abouse

kenjiew:
follow up from #35315 
You can see in that issue, I can't provide any evidence here, sorry.


kenjiew:
hi @ykk4ga ok we will delete it after this, please be patient, do you have more than 1 domain?, subdomain maybe?, Oh yeah, don't make a pull request, okay, we're processing it in this PR.

satr14washere (approved):

PR #35446, opened 2026-03-30
Delete domains/andradenathan.json

kenjiew:
see issue #35445 

Hi @andradenathan do you still need this?, and please confirm whether this is your action or not, this domain will be deleted, if you still need it such as for a redirect url or your website?


iostpa (approved):
User can always re-register the domain, though it would also be recommended to verify the domain as well.

PR #36171, opened 2026-04-11
Delete domains/q.anw.json

iostpa:
sorry @SubhrajitSain but i checked with orangc and the website isn't fine


iostpa:
i didnt sign up to use copilot :wilted_flower:

PR #37139, opened 2026-04-27
Add bdtools.is-a.dev

ashik347:
- [x] I have read and agree to the terms
- [x] I understand that this domain will be publicly available
- [x] I am not using this for illegal purposes
- [x] I understand that abuse may result in removal

Domain: bdtools.is-a.dev

### DIFF
--- a/domains/bdtools.json
+++ b/domains/bdtools.json
@@ -1,0 +1,9 @@
+{
+  "owner": {
+    "username": "ashik347",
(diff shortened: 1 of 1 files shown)


STICKnoLOGIC:
Closing;
- not dev related, means we need a "software dev" related portfolio/```personal``` site for subdomain and all ```projects``` must be in nested subdomain

PR #37656, opened 2026-05-06
travelguide.is-a.dev

Phw-nii:
A web portal for the TravelGuide project, where admins manage points of interest and users can download the latest Android APK via a stable QR code link


iostpa:
Same reason as #37658

PR #37751, opened 2026-05-08
Update preciousatwiine.is-a.dev - Add TXT verification record

PreciousAtwiine:
Adding Vercel TXT verification record to connect domain to my Vercel project.


STICKnoLOGIC (commented):
this is not the proper way to add vercel verification, you need to create another json file e.g. _vercel.yourdomain.json
then add "ALL" TXT records there including the txt record for  'www' nested subdomain

PR #38207, opened 2026-05-13
Register lcchang.is-a.dev

lcchang224:
Personal subdomain for self-hosted services. NS delegation to Cloudflare for management.

### DIFF
--- a/domains/lcchang.json
+++ b/domains/lcchang.json
@@ -3,7 +3,7 @@
       "username": "lcchang224",
       "email": "lcchang224@gmail.com"
     },
-    "record": {
+    "records": {
       "NS": [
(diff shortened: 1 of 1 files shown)


domains/lcchang.json
```diff
@@ -0,0 +1,12 @@
+ {
+    "owner": {
+      "username": "lcchang224",
+      "email": "lcchang224@gmail.com"
+    },
+    "record": {
```
Yunexiz: Change this to "records".

Yunexiz (changes requested):
You'll need to:
- Add a preview;
- Go through and check all the requirements (follow the PR template);
- State your reasons for requiring NS instead of other records (restricted access).

PR #38566, opened 2026-05-18
Add tools.is-a.dev

kaijunior2002:
Hi! Requesting tools.is-a.dev for a personal tools website running on a Cloudflare tunnel.


dragsbruh (changes requested):
- thats not how you use cf tunnels, you need ns records
- you arent eligible for ns records

PR #38697, opened 2026-05-20
fix amithnal.is-a.dev - disable Cloudflare proxy (522 error)

amiiboii:
Hi! Really sorry for the extra PR - my domain amithnal.is-a.dev was 
approved earlier today (PR #38683) but is returning a 522 error.

The issue is that Cloudflare's proxy conflicts with Tailscale Funnel, 
which handles its own HTTPS termination. Adding "proxied": false fixes it.

Only change is the single addition of "proxied": false to the existing file.
Really appreciate the earlier approval and sorry for the inconvenience!

### DIFF
--- a/domains/amithnal.json
+++ b/domains/amithnal.json
@@ -5,5 +5,6 @@
   },
   "records": {
     "CNAME": "server.taile34ada.ts.net"
-  }
+  },
+  "proxied": false
 }


dragsbruh (changes requested):
thats not the issue. proxy is turned off by default.

for funnels you cant use custom domains this way if i remember correctly. consider using a different hosting/proxy service. 

im not entirely sure so make sure to read the docs, you might be able to get custom domains for funnels another way if you have premium or something.

another note: you cannot cloudflare tunnels either, they need NS records and you arent eligible for them

amiiboii:
Hi! Really sorry for the confusion and the extra noise , I should have read the docs more carefully before opening this. Thank you so much for taking the time to explain, genuinely appreciate it! I'll close this and sort out a proper hosting solution. Thanks again for the help! 😊

PR #38751, opened 2026-05-20
ci(workflows): pin github actions to commit hashes

DEV-DIBSTER:
Updated all GitHub Actions dependencies across workflows to use full commit SHAs instead of mutable version tags (e.g., `actions/checkout`, `actions/setup-node`, and `actions/stale`). 

This hardens the CI/CD pipeline by ensuring action dependencies are strictly immutable and prevents potential supply chain attacks.

Seems like GitHub actions are by design insecure, starting to slowly lock down these insecure Actions CI/CD.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
         name: Tests
         runs-on: ubuntu-latest
         steps:
-            - uses: actions/checkout@v6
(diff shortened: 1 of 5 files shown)


iostpa (approved):
LGTM, though will it also be worth bumping the CI versions (ava and fs-extra)? I think the newer GitHub Action versions prefer Node 24.

wdhdev (approved):

PR #38825, opened 2026-05-21
Edit lemonfellow.json — disable proxy

l1203012:
## Description

Setting `"proxied": false` on `lemonfellow.is-a.dev` so it resolves directly to GitHub Pages instead of being proxied through Cloudflare.

## Reason

The subdomain hosts a personal portfolio at `l1203012.github.io/portfolio_v26` configured as a GitHub Pages custom domain. With Cloudflare proxying enabled, GitHub Pages cannot complete the Let's Encrypt ACME challenge, so the **Enforce HTTPS** option remains greyed out with the message *"Unavailable for your site because your domain is not properly configured to support HTTPS."*

Disabling the proxy allows GitHub to issue and serve a valid TLS certificate for the subdomain.

## Checklist

- [x] Edited only my own subdomain JSON
- [x] JSON is valid
- [x] Change is minimal (added single `"proxied": false` field)


dragsbruh (changes requested):
- proxy is disabled by default
- this is your first pr, wdym edit? the issue was you didnt even get lemonfellow.is-a.dev
- why did you remove pr template? readd it properly with checklist

create fresh pr and do it properly

PR #39533, opened 2026-05-28
feat: add professor-domains.is-a.dev

profx-ctrl:
Requesting professor-domains.is-a.dev for my free domain platform project


dragsbruh (changes requested):
sites must be software development related, also i dont think thats allowed

PR #39653, opened 2026-05-30
Register prism.is-a.dev

uxwangy-code:
Registering prism.is-a.dev (CNAME to a Netlify site). Personal knowledge hub. Thanks!


iostpa (changes requested):
We don't allow websites that are locked behind a login system.
Also we don't allow no-reply emails, you can simply remove the email line.

iostpa:
Forgot to include incomplete pull request as well, you need to fill out our pull request template.

PR #40018, opened 2026-06-04
Add JSON configuration for GitHub Pages challenge

MrUnknown-47:
# Requirements

- [x] I **agree** to the [Terms of Service](https://is-a.dev/terms). 
- [x] My file is following the [domain structure](https://docs.is-a.dev/domain-structure/). 
- [x] My website is **reachable** and **completed**. 
- [x] My website is **software development** related. 
- [x] My website is **not for commercial use**. 
- [x] I have provided contact information in the `owner` key. 
- [x] I have provided a preview of my website below. 

# Website Preview

https://mrunknown-47.github.io/portfolio/
Screenshot 2026-06-04 at 1 32 33 PM

# Website Purpose

It is a Portfolio Website.

PR #40137, opened 2026-06-05
Delete domains/vaibhav.json

iostpa:
@MrUnknown-47 Your domain got taken over by someone, so the domain is currently hosting a scam website. We are taking down this domain because of that.

Please re-register with the file for the root subdomain (vaibhav.is-a.dev), but also a valid file (you need to fix that file) from #40018 so that you can verify the domain.

### DIFF
--- a/domains/vaibhav.json
+++ b/domains/vaibhav.json
@@ -1,9 +1,0 @@
-{
-    "owner": {
-        "username": "MrUnknown-47",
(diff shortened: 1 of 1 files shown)


DEV-DIBSTER (approved):

iostpa:
Don't know why I had it on draft

PR #40333, opened 2026-06-08
Register fuckpaper.is-a.dev

xintiaoobao:
### Domain Registration

- **Subdomain**: fuckpaper.is-a.dev
- **Record Type**: CNAME
- **Target**: xintiaoobao.github.io


dragsbruh (changes requested):
against tos most likely, and pr template was removed

PR #40344, opened 2026-06-08
Register capybarabot.is-a.dev

and2311666:
Registering capybarabot.is-a.dev with an A record pointing to my server. Thanks!


dragsbruh (changes requested):
pr template was removed

PR #40897, opened 2026-06-15
add tatva.is-a.dev

tatvadesai:
Register `tatva.is-a.dev` as a CNAME to `tatva.tatvadesai.workers.dev`.


dragsbruh (changes requested):
- thats not how you use cf workers/tunnels
- they need ns records
- you need a valid reason for ns records
- prs without valid reason will be closed

PR #41542, opened 2026-06-22
Add email DNS records for eunsunglee.is-a.dev (Resend DKIM/SPF/DMARC)

eunsunglee0524:
# Requirements

- [x] I **agree** to the [Terms of Service](https://is-a.dev/terms).
- [x] My file is following the [domain structure](https://docs.is-a.dev/domain-structure/).
- [x] My website is **reachable** and **completed**.
- [x] My website is **software development** related.
- [x] My website is **not for commercial use**.
- [x] I have provided contact information in the `owner` key.
- [x] I have provided a preview of my website below.

# Website Preview

https://eunsunglee.is-a.dev

# Website Purpose

Adding email DNS records (DKIM, SPF, MX, DMARC) for transactional email 
via Resend to my existing domain eunsunglee.is-a.dev. 
These subdomains (send, _dmarc, resend._domainkey) are required 
for email sending verification and are not user-facing websites.

PR #41733, opened 2026-06-25
Delete domains/eunsunglee.json

notamitgamer:
@eunsunglee0524

We're terminating your subdomain `eunsunglee.is-a.dev` since it only says `hi` and doesn't have an actual project or anything software development related on it.

PR: #31105, #41542 

If you actually want to use it and want to appeal, just open an issue on the repo or drop us a message on Discord.

Proof: 
image


notamitgamer:
Closing as user fixed the issue in #41542

PR #42454, opened 2026-07-02
Register voti.is-a.dev

LimChangHun81:
Registering voti.is-a.dev as a CNAME to a Cloudflare Tunnel for a personal project.


LimChangHun81:
Closing — cfargotunnel.com CNAMEs aren't allowed per policy, using a different setup instead.

PR #42563, opened 2026-07-04
Register audiolingua.is-a.dev

Aliahasan399:
Adding audiolingua.is-a.dev → HF Spaces


MKishoreDev:
Replace `record` with `records`

PR #42717, opened 2026-07-05
Update webora.is-a.dev to Cloudflare NS records

talha50819:
Switching webora.is-a.dev from a CNAME to Netlify to NS delegation via Cloudflare (clint.ns.cloudflare.com, ophelia.ns.cloudflare.com). Cloudflare zone is set up with matching CNAME records pointing to webora-398.netlify.app so the site continues resolving to the same origin.


notamitgamer:

PR #42865, opened 2026-07-07
Add Resend DNS records for mail.jordi

jordig20:
Adds DNS records for Resend email sending on mail.jordi.is-a.dev.\n\nRecords included:\n- DKIM TXT for resend._domainkey.mail.jordi\n- SPF TXT and MX return-path records for send.mail.jordi\n- DMARC TXT for _dmarc.mail.jordi


Yunexiz (changes requested):
You need to first create mail.jordi.json, before creating send.mail.jordi.json.

PR #44216, opened 2026-07-19
Register nejatino.is-a.dev

omidadli:
Registering nejatino.is-a.dev with a CNAME record pointing to Vercel (cname.vercel-dns.com) for an animal rescue charity website.


Yunexiz:
Unfortunately, I do not think an animal rescue charity can be considered dev related. I must reject this PR. You can instead make a different site for the root subdomain, then use a nested subdomain for the animal rescue site.

PR #44418, opened 2026-07-21
Add wemesh.json for design8492

mdrakib52:
Adding domain wemesh.is-a.dev for design8492

### DIFF
--- a/domains/wemesh.json
+++ b/domains/wemesh.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "design8492",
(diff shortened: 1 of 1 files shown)


domains/wemesh.json
```diff
@@ -0,0 +1,9 @@
+{
+    "owner": {
+        "username": "design8492",
```
MaskDuck, line 3: please register using the account `design8492`

MaskDuck (changes requested):

PR #44601, opened 2026-07-22
Register erkatooy.is-a.dev

vengeoff:
New subdomain registration — mobile app backend for Erkatooy ERP (A record to a self-managed VPS).


vengeoff:
Closing — decided to go with plain HTTP on our own server instead of a proxied domain for now. Thanks!

PR #44788, opened 2026-07-24
Update _vercel.cibik.json

cibi-k:
<## Description
Updated the Vercel TXT verification token for `cibik.is-a.dev` after the previous token expired.

This PR only updates the verification record.


dragsbruh (approved):
you might want to change the domain type in vercel for `cibik.is-a.dev` from `redirect to another domain` to `connect to environment: production`

PR #44865, opened 2026-07-24
fix(kulindupr): migrate to records schema, add www + Vercel verification

kulindu24:
## Summary
Fixes `kulindupr.is-a.dev` so Vercel domain verification succeeds and adds `www` support.

- `domains/kulindupr.json` was using the deprecated singular `record` key (now a blocked field per `tests/json.test.js`), so `dnsconfig.js` (which reads `data.records.*`) never generated any actual DNS records for this domain. Migrated to `records`, with `A` as an array.
- Split Vercel's TXT ownership-verification value out into its own `_vercel.kulindupr.json` file, matching the convention used by other Vercel-linked domains (e.g. `_vercel.amartya.json`).
- Added `www.kulindupr.json` (CNAME to Vercel) and `_vercel.www.kulindupr.json` (TXT verification for `www`), since neither existed before.

## Test plan
- [x] Verified all 4 files parse as valid JSON and match the required/optional field rules in `tests/json.test.js` (required `owner`/`records` objects, no blocked fields, valid owner email, file naming rules).
- [x] Verified record values against `tests/records.test.js` (A is an array of valid public IPv4, CNAME passes hostname regex and isn't in `disallowed-cnames.json`, TXT is a string).
- [x] Compared structure 1:1 against existing merged examples (`amartya.json` / `_vercel.amartya.json` / `www.amartya.json` / `_vercel.www.amartya.json`).

### DIFF
--- a/domains/_vercel.kulindupr.json
+++ b/domains/_vercel.kulindupr.json
@@ -4,8 +4,6 @@
     "email": "kulindupr@gmail.com"
   },
   "records": {
-    "A": [
-      "216.198.79.1"
-    ]
+    "TXT": "vc-domain-verify=kulindupr.is-a.dev,908b433c019e4f0e11bc"
   }
 }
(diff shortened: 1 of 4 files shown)


ok-coder1:
This is not how you verify subdomains. Please read the docs.

PR #44937, opened 2026-07-25
ci: stop rebuilding the DNSControl docker action on every PR

### LINKED ISSUE #44795: unnecessary building of dnscontrol container in vaidation ci
Image

currently for tests on pull requests the docker image for is-a-dev/dnscontrol-action and the environment in general (fetching NPM packages) is rebuilt every time its run, totaling 13 seconds

the built image isnt even used for over 99.9% of PRs, which makes it time consuming and just unnecessary.

moneytosms:
Closes #44795.

This is an infrastructure change, so the domain checklist in the PR template does not apply (no website preview, no `domains/` file).

## Problem

GitHub builds docker-based actions during `Set up job`, before step-level `if:` conditions are evaluated. `is-a-dev/dnscontrol-action` declares `image: 'Dockerfile'`, so the image gets built on every single pull request.

The DNSControl step is gated on `dnsconfig.js` being in the changed files, which never happens on an ordinary domain pull request. So the build is pure waste. From run [30156964836](https://github.com/is-a-dev/register/actions/runs/30156964836):

```
Build is-a-dev/dnscontrol-action@main  11:51:58 -> 11:52:04   6s
Run actions/checkout@df4cb1c              :52:04 ->   :52:06   2s
Run npm install                           :52:06 ->   :52:11   5s
Fetch PR information                      :52:11 ->   :52:13   2s
Run DNSControl tests                      :52:13 ->   :52:13   0s   <- skipped
Run tests                                 :52:13 ->   :52:17   4s
```

Six of twenty seconds went to building an image for a step that did not run.

## Changes

**Download the DNSControl binary instead of using the docker action.**

For the `check` subcommand the action's `entrypoint.sh` reduces to `dnscontrol check --config <file>`. It explicitly skips `--creds` for `check`, and its `output` / `preview_comment` outputs are not consumed here (the step has no `id`), so nothing is lost. `DNSCONTROL_VERSION` is pinned to `4.14.0` to match the action, with a comment noting the two should be bumped together.

`publish.yml` still uses the action for `push` with Cloudflare credentials and is untouched.

**Commit a lockfile and use `npm ci`.**

There was no `package-lock.json`, so `npm install` resolved from the registry on every run. This is about reproducible dependency resolution, not speed. See the measurements below, where the time difference is negligible.

`actions/setup-node` is pinned to `24`, which is what `ubuntu-latest` already provides today, so this is not a version change. The suite was verified on both 22 and 24.

## Measurements

Run on `ubuntu` containers against this repository (13,857 domain files):

```
download + extract dnscontrol:   1s   (42MB binary)
dnscontrol check:                3s   (parses 16,359 records)

npm install (no lockfile):       2s
npm ci (warm npm cache):         1s
npm ci (cold npm cache):         2s
```

So the docker build is traded for a 1s download. The 3s `check` cost is unchanged, it was previously inside the action.

| | before | after |
|---|---|---|
| domain pull request (check skipped) | 6s wasted | 0s |
| `dnsconfig.js` pull request | 6s build + 3s check | 1s download + 3s check |

A typical domain pull request should go from about 20s to about 14s. Worth noting that the 10s figure in #44795 is not reachable: checkout, PR info and the ava suite are about 8s of irreducible work on top of job setup.

## On using a prebuilt container

#44795 suggests publishing a prebuilt image. That would still be pulled during `Set up job` for a step that is almost always skipped, so it would cost a couple of seconds where removing the action costs nothing. It also needs two separate merges in `is-a-dev/dnscontrol-action` (publish the image first, then flip `action.yml`), since pointing `image:` at an image that does not exist yet would break CI.

Publishing the image is still worthwhile for `publish.yml`, which does need the action. Happy to open that separately if you want it.

## Verification

Ran the job steps in `node:22-bookworm` and `node:24-bookworm` containers with this repository mounted:

- `npm ci` from the committed lockfile installs 186 packages
- binary download and `tar xz dnscontrol` work, the binary is at the archive root
- `dnscontrol check --config dnsconfig.js` reports `No errors.` and exits 0
- a deliberately broken `dnsconfig.js` exits non-zero, so the step still fails CI correctly
- `dnscontrol print-ir` confirms the check is not vacuous: 1 zone, 16,359 records parsed from 13,857 files
- `dnscontrol` resolves the `glob("./domains")` in `dnsconfig.js` relative to the config file rather than the working directory, so dropping the action's `cd "$(dirname "$CONFIG_ABS_PATH")"` is safe
- `npx ava tests/*.test.js` passes 13/13 on both node 22 and 24
- `actionlint` reports no problems on any workflow

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,12 @@ jobs:
         steps:
             - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
 
-            - run: npm install
+            - uses: actions/setup-node@249970729cb0ef3589644e2896645e5dc5ba9c38 # v6.5.0
+              with:
+                  node-version: 24
+                  cache: npm
+
+            - run: npm ci
 
             - name: Fetch PR information
               if: github.event_name == 'pull_request'
@@ -42,11 +47,17 @@ jobs:
               env:
                   GH_TOKEN: ${{ github.token }}
 
+            # Docker actions are built during "Set up job" even when their step is
+            # skipped, so using is-a-dev/dnscontrol-action here costs ~6s on every PR
+            # for a check that almost never runs. Download the binary instead.
+            # Keep DNSCONTROL_VERSION in sync with is-a-dev/dnscontrol-action.
             - name: Run DNSControl tests
               if: github.event_name == 'pull_request' && contains(fromJson(env.CHANGED_FILES), 'dnsconfig.js')
-              uses: is-a-dev/dnscontrol-action@main
-              with:
-                  args: check
+              env:
+                  DNSCONTROL_VERSION: 4.14.0
+              run: |
+                  curl -fsSL "https://github.com/DNSControl/dnscontrol/releases/download/v${DNSCONTROL_VERSION}/dnscontrol_${DNSCONTROL_VERSION}_linux_amd64.tar.gz" | tar xz dnscontrol
+                  ./dnscontrol check --config dnsconfig.js
 
             - name: Run tests
               run: npx ava tests/*.test.js --timeout=1m
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,2380 @@
+{
+  "name": "register",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "devDependencies": {
+        "ava": "^6.2.0",
+        "fs-extra": "^11.2.0"
+      }
+    },
+    "node_modules/@isaacs/cliui": {
+      "version": "8.0.2",
+      "resolved": "https://registry.npmjs.org/@isaacs/cliui/-/cliui-8.0.2.tgz",
+      "integrity": "sha512-O8jcjabXaleOG9DQ0+ARXWZBTfnP4WNAqzuiJK7ll44AmxGKv/J2M4TPjxjY3znBCfvBXFzucm1twdyFybFqEA==",
+      "dev": true,
+      "license": "ISC",
+      "dependencies": {
+        "string-width": "^5.1.2",
+        "string-width-cjs": "npm:string-width@^4.2.0",
+        "strip-ansi": "^7.0.1",
+        "strip-ansi-cjs": "npm:strip-ansi@^6.0.1",
+        "wrap-ansi": "^8.1.0",
+        "wrap-ansi-cjs": "npm:wrap-ansi@^7.0.0"
+      },
+      "engines": {
+        "node": ">=12"
+      }
+    },
+    "node_modules/@isaacs/cliui/node_modules/emoji-regex": {
+      "version": "9.2.2",
+      "resolved": "https://registry.npmjs.org/emoji-regex/-/emoji-regex-9.2.2.tgz",
+      "integrity": "sha512-L18DaJsXSUk2+42pv8mLs5jJT2hqFkFE4j21wOmgbUqsZ2hL72NsUU785g9RXgo3s0ZNgVl42TiHp3ZtOv/Vyg==",
+      "dev": true,
+      "license": "MIT"
+    },
+    "node_modules/@isaacs/cliui/node_modules/string-width": {
+      "version": "5.1.2",
+      "resolved": "https://registry.npmjs.org/string-width/-/string-width-5.1.2.tgz",
+      "integrity": "sha512-HnLOCR3vjcY8beoNLtcjZ5/nxn2afmME6lhrDrebokqMap+XbeW8n9TXpPDOqdGK5qcI3oT0GKTW6wC7EMiVqA==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "eastasianwidth": "^0.2.0",
+        "emoji-regex": "^9.2.2",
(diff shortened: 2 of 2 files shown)


moneytosms:
This pull request's own CI run gives real numbers, and they are better than the estimate in the description. I said about 14s, the actual result is 12s.

Before, run [30156964836](https://github.com/is-a-dev/register/actions/runs/30156964836) on `main`:

```
11:51:57 -> 11:51:58  Set up job
11:51:58 -> 11:52:04  Build is-a-dev/dnscontrol-action@main    6s
11:52:04 -> 11:52:06  Run actions/checkout                     2s
11:52:06 -> 11:52:11  Run npm install                          5s
11:52:11 -> 11:52:13  Fetch PR information                     2s
11:52:13 -> 11:52:13  Run DNSControl tests            skipped  0s
11:52:13 -> 11:52:17  Run tests                                4s
                                                       total  20s
```

After, run [30157968229](https://github.com/is-a-dev/register/actions/runs/30157968229) on this branch:

```
12:25:57 -> 12:25:58  Set up job
12:25:58 -> 12:26:00  Run actions/checkout                     2s
12:26:00 -> 12:26:00  Run actions/setup-node                   0s
12:26:00 -> 12:26:03  Run npm ci                               3s
12:26:03 -> 12:26:04  Fetch PR information                     1s
12:26:04 -> 12:26:04  Run DNSControl tests            skipped  0s
12:26:04 -> 12:26:08  Run tests                                4s
                                                       total  12s
```

20s to 12s, so about 40% off. The `Build is-a-dev/dnscontrol-action` step is gone entirely rather than shortened, which was the point.

Two things I had guessed conservatively and got wrong in the description:

- `actions/setup-node` cost 0s, not the 1s to 2s I budgeted for it.
- This run had a cold npm cache, since it is the first run on the branch and there was nothing to restore yet. `npm ci` still came in at 3s against the previous `npm install` at 5s. Later runs should hit the cache and take a little less.

Everything else in the description holds. Happy to rebase or adjust if you would rather see this split into two pull requests, one for the docker action and one for the lockfile.

.github/workflows/ci.yml
```diff
@@ -42,11 +47,17 @@ jobs:
...
+            # Keep DNSCONTROL_VERSION in sync with is-a-dev/dnscontrol-action.
             - name: Run DNSControl tests
               if: github.event_name == 'pull_request' && contains(fromJson(env.CHANGED_FILES), 'dnsconfig.js')
-              uses: is-a-dev/dnscontrol-action@main
-              with:
-                  args: check
+              env:
+                  DNSCONTROL_VERSION: 4.14.0
+              run: |
+                  curl -fsSL "https://github.com/DNSControl/dnscontrol/releases/download/v${DNSCONTROL_VERSION}/dnscontrol_${DNSCONTROL_VERSION}_linux_amd64.tar.gz" | tar xz dnscontrol
```
dragsbruh, line 59: we use a fork of dnscontrol for a reason, this is not exactly a one pr change

.github/workflows/ci.yml
```diff
@@ -42,11 +47,17 @@ jobs:
...
+            # skipped, so using is-a-dev/dnscontrol-action here costs ~6s on every PR
+            # for a check that almost never runs. Download the binary instead.
+            # Keep DNSCONTROL_VERSION in sync with is-a-dev/dnscontrol-action.
             - name: Run DNSControl tests
               if: github.event_name == 'pull_request' && contains(fromJson(env.CHANGED_FILES), 'dnsconfig.js')
-              uses: is-a-dev/dnscontrol-action@main
-              with:
-                  args: check
+              env:
+                  DNSCONTROL_VERSION: 4.14.0
```
dragsbruh, line 57: the latest version of upstream dnscontrol is 4.44.1 from what i see

.github/workflows/ci.yml
```diff
@@ -23,7 +23,12 @@ jobs:
         steps:
             - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
 
-            - run: npm install
+            - uses: actions/setup-node@249970729cb0ef3589644e2896645e5dc5ba9c38 # v6.5.0
+              with:
+                  node-version: 24
+                  cache: npm
+
+            - run: npm ci
```
dragsbruh, line 31: the issue also covers npm packages but thats a minor difference

.github/workflows/ci.yml
```diff
@@ -23,7 +23,12 @@ jobs:
         steps:
             - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
 
-            - run: npm install
+            - uses: actions/setup-node@249970729cb0ef3589644e2896645e5dc5ba9c38 # v6.5.0
+              with:
+                  node-version: 24
+                  cache: npm
+
```
dragsbruh, line 30: this entire step is unnecessary, if you manually checked ci logs you would see that

dragsbruh (changes requested):

dragsbruh:
please do not use ai

.github/workflows/ci.yml
dragsbruh: the fork uses 4.14.0 so this difference is okay i guess

https://github.com/is-a-dev/dnscontrol-action/blob/main/Dockerfile

PR #45198, opened 2026-07-27
Redirect divyakush.is-a.dev to www.divyakush.com (URL record)

Divyakush2006:
I've moved my portfolio to a new primary domain, **www.divyakush.com**. This changes my existing `divyakush.is-a.dev` entry to a native `URL` redirect pointing there, so the old subdomain permanently forwards to the new site. Ownership (Divyakush2006) is unchanged.

### DIFF
--- a/domains/divyakush.json
+++ b/domains/divyakush.json
@@ -4,6 +4,6 @@
     "email": "divyakushpunjabi@gmail.com"
   },
   "records": {
-    "CNAME": "divyakush2006.github.io"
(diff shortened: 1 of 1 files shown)


Divyakush2006:
No longer needed — I'm handling the redirect to www.divyakush.com via GitHub Pages on the existing CNAME, which is already live. Keeping my current divyakush.is-a.dev record unchanged. Thanks!

PR #45571, opened 2026-07-31
Revert "Add server configuration for FahadNaeem424"

notamitgamer:
Reverts is-a-dev/register#45490

abused, vpn service
related: #45493


notamitgamer:
my bad, i thought this sub also will be used for vpn service.

PR #45989, opened 2026-08-04
Add puter-batch-2026.is-a.dev (NS+DS delegate to deSEC)

wxlei2004:
## Domain
`puter-batch-2026.is-a.dev`

## Records
- **NS** → `ns1.desec.io`, `ns2.desec.org` (delegate zone to deSEC)
- **DS** → DNSSEC key tag 41448 (algorithm 13, digest types 2 and 4)

## Purpose
Self-hosted email domain verification (MX + TXT live on deSEC after NS delegation).

## Checklist
- [x] File is lowercase and in `domains/`
- [x] Uses `records` (not `record`)
- [x] NS + DS combination only (valid per tests)
- [x] CI Tests passed

### DIFF
--- a/domains/puter-batch-2026.json
+++ b/domains/puter-batch-2026.json
@@ -1,0 +1,14 @@
+{
+  "owner": {
+    "username": "wxlei2004"
+  },
+  "record": {
+    "MX": ["smtp.215.im"],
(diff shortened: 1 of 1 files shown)


notamitgamer:

PR #46222, opened 2026-08-05
sipir.is-a.dev

sipircip:
Registering sipir.is-a.dev with NS records for Cloudflare DNS.


Stef-DP:
domain does not use PR template, does not provide reasoning for NS (your description says you want NS) and does not add any file

PR #46223, opened 2026-08-05
Add platform.kozlov.json configuration file

dryn85-max:
Purpose

Create the parent subdomain platform.kozlov.is-a.dev required for subsequent ACM DNS validation records.

This is a temporary placeholder record that will be updated to the actual API Gateway target after the AWS custom domain is configured.


wdhdev:
Put the PR template back and open a new PR

PR #46287, opened 2026-08-06
Register serviceos.is-a.dev

hasangonen91:
Registering serviceos.is-a.dev for serviceOS — a free online appointment system for barbershops and beauty salons in Turkey.

CNAME: customer-web.mkizil916.workers.dev


hasangonen91:
workers.dev CNAME not allowed by policy; switching to us.kg free domain via Cloudflare

PR #46587, opened 2026-08-08
Register gpt-bridge.is-a.dev

marona190:
CNAME record pointing to a personal Cloudflare Tunnel used for a private local dev bridge. No public content is served.


marona190:
Closing this — I've moved the endpoint to a different ingress that doesn't need an is-a.dev record, so this registration is no longer needed. Sorry for the noise, and thanks for maintaining the registry.

PR #46870, opened 2026-08-11
Add wekathana.is-a.dev

giuseppeoliveira:
Register WeKathana subdomain pointing to kathana.web3-gdev.workers.dev


giuseppeoliveira:
Closing — will keep using the Cloudflare Workers URL instead of an is-a.dev subdomain.

PR #48560, opened 2026-08-23
Add dsl.is-a.dev

prayasabhinav:
**Subdomain:** `dsl.is-a.dev`
**Record:** CNAME → `dsl-status.aiclub.anu.edu.in`

A course tool for sixty-three interaction-design students at Anant National University. Each writes one line a day against the day's question and reads what the others wrote. Self-hosted on our own server behind our own reverse proxy — not Vercel or Netlify, so no TXT verification is needed.

The short name is so students can type it from memory.

### DIFF
--- a/domains/dsl.json
+++ b/domains/dsl.json
@@ -1,0 +1,9 @@
+{
+    "owner": {
+        "username": "prayasabhinav",
+        "email": "me@prayasabhinav.net"
+    },
+    "records": {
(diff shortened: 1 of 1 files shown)


prayasabhinav:
Withdrawing — on reflection the name reads as a claim about the project that doesn't fit a university course tool. Thanks for running this, it's clearly the best-maintained of these services.

PR #48593, opened 2026-08-23
Update jifeng.is-a.dev DNS to CNAME

Yangchengen-hub:
Update DNS record from A record to CNAME pointing to Cloudflare Pages (jifeng-cn.pages.dev) for better CDN and HTTPS support.


gameroman (changes requested):
Please use the PR template

PR #48701, opened 2026-08-23
add test for TXT record when vercel and remove invalid files

gameroman:
Added a test for invalid TXT records and removed invalid ones


tests/records.test.js
```diff
@@ -238,6 +238,15 @@ function validateRecordValues(t, data, file) {
             const values = Array.isArray(value) ? value : [value];
             values.forEach((record, idx) => {
                 t.true(typeof record === "string", `${file}: TXT record value should be a string at index ${idx}`);
+
+                if (record.startsWith("vc-domain-verify=")) {
+                    const domainToken = record.split("vc-domain-verify=")[1].split(",")[0];
+
+                    t.true(
+                        domainToken.endsWith(".is-a.dev"),
```
gameroman: I've kept it simple just to catch cases like this, but it could be more strict if needed

https://redirect.github.com/is-a-dev/register/pull/48694/changes#diff-eeed588888634fd436b0f3bf96619cf5f5159a42a5f662f0c2096889024bd66aR7

dragsbruh (dismissed):
can you also remove any parent domains of these domains? only if they also are configured to use vercel ofc

make sure that the they didnt change it to a diff hosting provider later, in which case remove only the `_vercel.` file and leave parent untouched

would also be a good idea to check if the vercel record is in `_vercel.domain.json` and not `domain.json` or `_vercel.sub.domain.json` or stuff like that

would also recommend pinging owners of any domains you are removing

gameroman:
> can you also remove any parent domains of these domains? only if they also are configured to use vercel ofc
> 
> make sure that the they didnt change it to a diff hosting provider later, in which case remove only the `_vercel.` file and leave parent untouched
> 
> would also recommend pinging owners of any domains you are removing

I think that's out of scope for this PR currently

> would also be a good idea to check if the vercel record is in `_vercel.domain.json` and not `domain.json` or `_vercel.sub.domain.json` or stuff like that

Yeah, I mentioned that in https://github.com/is-a-dev/register/pull/48701#discussion_r3839147272

I'd also do that in a separate PR probably

Lets see what @wdhdev thinks

wdhdev (changes requested):
Might be worth modifying the test so require them to be on _vercel nested subdomains & checking it somewhat matches the root subdomain, e.g. if it's on `_vercel.william.is-a.dev` the verification TXT(s) need to end with `.william.is-a.dev` or equal `william.is-a.dev` as nested subdomain Vercel verification records e.g. `blog.william.is-a.dev` go on `_vercel.william.is-a.dev` anyway (some people have arrays of TXT records). 

Essentially just verify it's on _vercel and that the TXT matches the subdomain or a nested subdomain.

gameroman:
> Might be worth modifying the test so require them to be on _vercel nested subdomains & checking it somewhat matches the root subdomain, e.g. if it's on `_vercel.william.is-a.dev` the verification TXT(s) need to end with `.william.is-a.dev` or equal `william.is-a.dev` as nested subdomain Vercel verification records e.g. `blog.william.is-a.dev` go on `_vercel.william.is-a.dev` anyway (some people have arrays of TXT records).
> 
> Essentially just verify it's on _vercel and that the TXT matches the subdomain or a nested subdomain.

I can do it in this PR or in a follow-up PR

wdhdev (approved):

PR #48965, opened 2026-08-25
Register byjiguel.is-a.dev

jigggle:
Registering byjiguel.is-a.dev for personal use


gameroman (changes requested):
Files should be in the `domains/` folder

PR #49570, opened 2026-08-28
Register locale.is-a.dev for email routing (Resend)

hanxsoloxx:
This domain (locale.is-a.dev) is used purely for email routing via Resend DKIM, SPF/MX, and DMARC records only. There's no website to preview, since this domain doesn't host any content.


gameroman:
Please fill in the PR template correctly

PR #50171, opened 2026-08-31
chore: add tests to block "is-a-dev" in subdomain names

### LINKED ISSUE #50094: ci: block “is-a-dev” in subdomain names
Block trailing and preceding “is-a-dev” references in subdomain names e.g. “william-is-a-dev.is-a.dev”

gameroman:
Closes #50094


wdhdev (changes requested):
https://github.com/is-a-dev/register/blob/7c692aeba5548ef0068c2333f6c83648bd8e8a9f/tests/json.test.js#L98

Probably easier to just do something like this ^
So it's altogether in the same file.

wdhdev (approved):

PR #52046, opened 2026-09-12
Register stock-tracker-v2.is-a.dev

Anqarium:
Registering stock-tracker-v2.is-a.dev to point to my Cloudflare zone (NS delegation) for a personal, self-hosted stock tracking app behind Cloudflare Tunnel.


Anqarium:
Closing — this subdomain is for a private, self-hosted personal app (behind auth), so I can't provide a public preview as required by the template. Will use a different domain instead. Sorry for the noise!

PR #53097, opened 2026-09-19
Fixing BOM 2nd try

notamitgamer:
fixing BOM 2nd try


notamitgamer:
Fixed from my side.
image

iostpa (approved):
Should be the fix now

DevMatei (approved):

